Issue #536, opened 2020-01-01
Payment Type: Edit Payment Type


abhaychawla:
**Goal:**
As an organization admin, I require the ability to edit a payment type as at,

https://demo.mifos.io/#/editPaymentType/1

![edit payment type](https://user-images.githubusercontent.com/16948598/71646956-e9433580-2cff-11ea-9fc6-1e13498fc998.PNG)

Resources:
https://mifosforge.jira.com/wiki/spaces/docs/pages/90243216/Payment+type
https://demo.mifos.io/api-docs/apiLive.htm#paymenttype_update

PR #560, opened 2020-01-09
feat: add edit payment type component

Rishabh106:
## Description
Added routing to the Edit Button in Payment types table, implemented the API and designed the UI.

#536 

## Screenshots, if any
edit-payment-type

## Checklist

### DIFF
--- a/src/app/organization/organization-routing.module.ts
+++ b/src/app/organization/organization-routing.module.ts
@@ -21,6 +21,7 @@ import { AdhocQueryComponent } from './adhoc-query/adhoc-query.component';
 import { ViewAdhocQueryComponent } from './adhoc-query/view-adhoc-query/view-adhoc-query.component';
 import { TellersComponent } from './tellers/tellers.component';
 import { PaymentTypesComponent } from './payment-types/payment-types.component';
+import { EditPaymentTypeComponent } from './payment-types/edit-payment-type/edit-payment-type.component';
 import { PasswordPreferencesComponent } from './password-preferences/password-preferences.component';
 import { EntityDataTableChecksComponent } from './entity-data-table-checks/entity-data-table-checks.component';
 import { WorkingDaysComponent } from './working-days/working-days.component';
@@ -38,6 +39,7 @@ import { AdhocQueriesResolver } from './adhoc-query/adhoc-queries.resolver';
 import { AdhocQueryResolver } from './adhoc-query/adhoc-query.resolver';
 import { TellersResolver } from './tellers/tellers.resolver';
 import { PaymentTypesResolver } from './payment-types/payment-types.resolver';
+import { PaymentTypeResolver } from './payment-types/payment-type.resolver';
 import { PasswordPreferencesTemplateResolver } from './password-preferences/password-preferences-template.resolver';
 import { EntityDataTableChecksResolver } from './entity-data-table-checks/entity-data-table-checks.resolver';
 import { WorkingDaysResolver } from './working-days/working-days.resolver';
@@ -171,6 +173,20 @@ const routes: Routes = [
               path: 'create',
               component: CreatePaymentTypeComponent,
               data: { title: extract('Create Payment Type'), breadcrumb: 'Create Payment Type'}
+            },
+            {
+              path: ':id',
+              data: { routeParamBreadcrumb: 'id', addBreadcrumbLink: false },
+              children: [
+                {
+                  path: 'edit',
+                  component: EditPaymentTypeComponent,
+                  data: { title: extract('Edit Payment Type'), breadcrumb: 'Edit', routeParamBreadcrumb: false },
+                  resolve: {
+                    paymentType: PaymentTypeResolver
+                  }
+                }
+              ]
             }
           ]
         },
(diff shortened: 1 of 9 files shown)


src/app/organization/payment-types/payment-types.component.html
```diff
@@ -48,7 +48,7 @@
       <ng-container matColumnDef="actions">
         <th mat-header-cell *matHeaderCellDef> Actions </th>
         <td mat-cell *matCellDef="let paymentType">
-          <button mat-button color="primary">
+          <button mat-button (click)="editPaymentType(paymentType.id)" color="primary">
```
punwai:
```suggestion
          <button mat-button [routerLink]="['edit' ,paymentType.id]" color="primary">
```

src/app/organization/payment-types/payment-types.component.ts
```diff
@@ -72,6 +73,14 @@ export class PaymentTypesComponent implements OnInit {
     this.dataSource.sort = this.sort;
   }
 
+  /**
+   * Redirects to edit payment type
+   * @param {string} paymentTypeId Payment Type ID of payment type to be edited.
+   */
+  editPaymentType(paymentTypeId: string) {
+    this.router.navigate([`edit/${paymentTypeId}`], { relativeTo: this.route });
```
punwai: This can be deleted.

src/app/organization/payment-types/payment-types.component.ts
```diff
@@ -42,6 +42,7 @@ export class PaymentTypesComponent implements OnInit {
    */
   constructor(private organizationService: OrganizationService,
               private route: ActivatedRoute,
+              private router: Router,
```
punwai: Delete this.

src/app/organization/payment-types/payment-types.component.ts
```diff
@@ -1,7 +1,7 @@
 /** Angular Imports */
 import { Component, OnInit, ViewChild } from '@angular/core';
 import { MatTableDataSource, MatPaginator, MatSort, MatDialog } from '@angular/material';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
```
punwai: Remove the import

punwai:
I am not a mentor, but I believe that an implementation along these lines would be much cleaner.

src/app/organization/payment-types/payment-types.component.html
Rishabh106: Thanks for the suggestion :)

src/app/organization/organization-routing.module.ts
```diff
@@ -171,6 +173,14 @@ const routes: Routes = [
               path: 'create',
               component: CreatePaymentTypeComponent,
               data: { title: extract('Create Payment Type'), breadcrumb: 'Create Payment Type'}
+            },
+            {
+              path: 'edit/:id',
```
punwai: It would probably be better if id comes before edit, to ensure consistency with all the other components. Also, it would be better to implement ':id' as a route on its own, and then have 'edit' as its child, as we will need this later to implement the viewPaymentTypes component. Push the data, resolve and component into the newly created 'edit' child route.

punwai: See system>account-number-preference as reference.

punwai: Regarding your query on gitter just now, see system>account-number-preference to see how you should deal with breadcrumbs.

punwai: I apologise for that. You are correct, we don't need viewPaymentType in this component. Wait for Abhay's comment on this but I'd add the ID into the bread crumbs so that it becomes `Edit #ID` as we can't have a separate ID in the breadcrumb which redirects to nothing.

punwai:
If you want to still apply the feature we discussed where the breadcrumb would look like: ##IdOrName## / Edit, you could still apply the same code mentioned where :id is the parent and edit is the child route. However, this time you can specify [addBreadcrumbLink] = false in the :id route, which would disable the user from clicking into an unimplemented page.

punwai:
I don't want to waste your time implementing something which may not be right, I'd wait for Abhay to comment, but here is what my opinion of what could be implemented for this is.

src/app/organization/payment-types/edit-payment-type/edit-payment-type.component.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Creates and sets the payment type form.
+   */
+  ngOnInit() {
+    this.createpaymentTypeForm();
+  }
+
+  /**
+   * Creates the payment type form.
+   */
+  createpaymentTypeForm() {
```
Jov03:
```suggestion
  createPaymentTypeForm() {
```

src/app/organization/organization-routing.module.ts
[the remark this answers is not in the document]
Rishabh106: can I use 'edit' as a child component of ':id' to display the required things?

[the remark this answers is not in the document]
punwai: As we've talked in gitter, I've had a look at your code and I believe that using [addBreadcrumbLink]=false like discussed here should do the job. I can't guarantee it's the best way for the mentors, but I'd personally give it a pass.

Issue #612, opened 2020-02-21
Bad UI: arrow of language select not aligned correctly


anirudhsingh20:
## Description
Arrow of language select not aligned correctly (vertically)

## Steps to Reproduce
1. Login
2. Check header in desktop view and see language select.

## Expected Behaviour
![mifos-lang-fix](https://user-images.githubusercontent.com/49402380/75049926-02db4f00-54f1-11ea-83c5-aa2ceef621b7.PNG)

## Actual Behaviour

Arrow has no margin-top and very close to the language text

![mifos-lang](https://user-images.githubusercontent.com/49402380/75049877-efc87f00-54f0-11ea-96c8-375f2886adfa.PNG)

## Settings
- Mifos X version: 0.0.0 development
- Browser used: Chrome

## Additional Context

I want to work on this issue .

gittysachin:
Hi @anirudhsingh20 
Are you still working on this issue? If not then I'd like to work on this issue.

anirudhsingh20:
yes .

ashutoshsenger:
I'm working on this issue now.

DivineWisdom:
Hello has this issue been fixed? if No i would like to work on it

ashutoshsenger:
yes it has been fixed, you can see pull request has been created already, so you can move to other issues.

abhaychawla:
Closing this issue as it is unnecessary, we are using Angular Material components and as we have not done any customization for `mat-select` this is the current default behavior. If there is an issue with `mat-select`, it can be reported at the official Angular Material repository.

PR #665, opened 2020-03-14
arrow of language select was not aligned now fixed

ashutoshsenger:
In Reference to  #612. 

I have fixed the wrong alignment of the arrow in the mifox-language-selector.

### DIFF
--- a/src/app/shared/language-selector/language-selector.component.scss
+++ b/src/app/shared/language-selector/language-selector.component.scss
@@ -1,3 +1,9 @@
 #language-selector {
   max-width: 4rem;
 }
+
+::ng-deep .mat-select-arrow {
+  color: rgb(10, 73, 99);
+  position: relative!important;
+  top: 8px!important;
+}


src/app/shared/language-selector/language-selector.component.scss
```diff
@@ -1,3 +1,9 @@
 #language-selector {
   max-width: 4rem;
 }
+
+::ng-deep .mat-select-arrow {
+  color: rgb(10, 73, 99);
+  position: relative!important;
+  top: 8px!important;
+}
```
karantakalkar, line 9:
```suggestion
::ng-deep .mat-select-arrow {
  position: relative;
  top: 8px;
}
```

karantakalkar: add new line at EOF

karantakalkar:
Abhay, I think it did fix the alignment well as i ran it on my machine, however yes i didn't realize ng deep depreciation.

abhaychawla:
It might be due to some other issue related to Angular Material. If its a issue related to the web app specifically, then we need to check if some other style that has been applied here is affecting the alignment of the arrow.

Issue #721, opened 2020-04-07
Has Permission Directive is required


karantakalkar:
Has Permission Directive is required to enforce permissions according to the logged-in user's role as in community-app.

![Screenshot from 2020-04-07 19-45-31](https://user-images.githubusercontent.com/54709463/78679908-7225ca80-7908-11ea-8650-5e0ae8a1e44c.png)

Directives Module needs to be Initialized

![Screenshot from 2020-04-07 19-48-19](https://user-images.githubusercontent.com/54709463/78680196-d3e63480-7908-11ea-9d45-777b7862dbf3.png)

karantakalkar:
@abhaychawla I'm on it

PR #722, opened 2020-04-07
feat: add has-permission structural directive

karantakalkar:
## Description
Initialized Directives Module and Developed Has-permission structural directive.
Refactored `accounting.component.html` to use the has-permission directive.

#721 

## Checklist

### DIFF
--- a/src/app/directives/has-permission/has-permission.directive.ts
+++ b/src/app/directives/has-permission/has-permission.directive.ts
@@ -1,0 +1,76 @@
+/** Angular Imports */
+import { Directive, Input, TemplateRef, ViewContainerRef } from '@angular/core';
+
+/** Custom Services */
+import { AuthenticationService } from '../../core/authentication/authentication.service';
+
+/**
+ * Has Permission Directive
+ */
+@Directive({
+  selector: '[mifosxHasPermission]'
+})
+export class HasPermissionDirective {
+
+  /** User Permissions */
+  private userPermissions: any[];
+
+  /**
+   * Extracts User Permissions from User Credentials
+   * @param {TemplateRef} templateRef Template Reference
+   * @param {ViewContainerRef} viewContainer View Container Reference
+   * @param {AuthenticationService} authenticationService AuthenticationService
+   */
+  constructor(private templateRef: TemplateRef<any>,
+              private viewContainer: ViewContainerRef,
+              private authenticationService: AuthenticationService) {
+    const savedCredentials = this.authenticationService.getCredentials();
+    this.userPermissions = savedCredentials.permissions;
+  }
+
+  /**
+   * Evaluates the condition to show template.
+   */
+  @Input()
+  set mifosxHasPermission(permission: any) {
+    if (typeof permission !== 'string') {
+      throw new Error('hasPermission value must be a string');
+    }
+    if (this.hasPermission(permission)) {
+      /** Shows Template */
+      this.viewContainer.createEmbeddedView(this.templateRef);
+    } else {
+      /** Hides Template */
+      this.viewContainer.clear();
+    }
+  }
+
+  /**
+   * Checks if user is permitted.
+   * @param {string} permission Permission
+   * @returns {true}
+   * -`ALL_FUNCTIONS`: user is a Super user.
+   * -`ALL_FUNCTIONS_READ`: user has all read permissions and passed permission is 'read' type.
+   * - User has special permission to access that feature.
+   * @returns {false}
+   * - Passed permission doesn't fall under either of above given permission grants.
+   * - No value was passed to the has permission directive.
+   */
+  private hasPermission(permission: string) {
+    permission = permission.trim();
+    if (this.userPermissions.includes('ALL_FUNCTIONS')) {
+      return true;
+    } else if (permission !== '') {
+        if (permission.substring(0, 5) === 'READ_' && this.userPermissions.includes('ALL_FUNCTIONS_READ')) {
+          return true;
+        } else if (this.userPermissions.includes(permission)) {
+          return true;
+        } else {
+          return false;
+        }
+    } else {
+      return false;
+    }
+  }
+
+}
--- a/src/app/accounting/accounting.component.html
+++ b/src/app/accounting/accounting.component.html
@@ -8,7 +8,7 @@
 
         <mat-nav-list>
 
-          <mat-list-item [routerLink]="['journal-entries/frequent-postings']">
+          <mat-list-item *mifosxHasPermission="'CREATE_JOURNALENTRY'" [routerLink]="['journal-entries/frequent-postings']">
             <mat-icon matListIcon>
               <fa-icon icon="sync" size="sm"></fa-icon>
             </mat-icon>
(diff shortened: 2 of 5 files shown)


karantakalkar:
build error due to timeout

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+  @Input()
+  set mifosxHasPermission(permission: any) {
+    if (typeof permission !== 'string') {
+      throw new Error('hasPermission value must be a string');
+    }
+    if (permission.charAt(0) === '!') {
+      this.isNegation = true;
+      permission = permission.slice(1);
+    }
+    if (this.hasPermission(permission) && !this.isNegation) {
```
abhaychawla, line 46: In community app we have two conditions here, did you check the specific use case for both possibilities? Why have we removed one of the conditions? https://github.com/openMF/community-app/blob/develop/app/scripts/directives/HasPermissionDirective.js

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+      this.viewContainer.createEmbeddedView(this.templateRef);
+      this.isVisible = true;
```
abhaychawla, line 48: Are you using this property anywhere?

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Checks if user is permitted.
+   * @param {string} permission Permission
+   * @returns {true} for 'all functions', 'all function read', and special user permissions.
+   * @returns {false} for null/undefined permission or lack of permission.
+   */
+  private hasPermission(permission: string) {
+    permission = permission.trim();
+    if (this.userPermissions.includes('ALL_FUNCTIONS')) {
+      return true;
+    } else if (permission !== null || undefined) {
```
abhaychawla, line 65:
```suggestion
    } else if (permission !== null || undefined) {
```
undefined is not compared with any parameter.

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Checks if user is permitted.
+   * @param {string} permission Permission
+   * @returns {true} for 'all functions', 'all function read', and special user permissions.
+   * @returns {false} for null/undefined permission or lack of permission.
+   */
+  private hasPermission(permission: string) {
+    permission = permission.trim();
+    if (this.userPermissions.includes('ALL_FUNCTIONS')) {
```
abhaychawla, line 63: Make sure userPermissions are available first (the array is not empty)

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+      this.viewContainer.clear();
+      this.isVisible = false;
+    }
+  }
+
+  /**
+   * Checks if user is permitted.
+   * @param {string} permission Permission
+   * @returns {true} for 'all functions', 'all function read', and special user permissions.
+   * @returns {false} for null/undefined permission or lack of permission.
```
abhaychawla, line 59: Comments are not appropriate, check and add proper explanation (reason why true is returned for each case)
https://github.com/openMF/community-app/blob/develop/app/scripts/controllers/main/MainController.js

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: No this is just informative variable

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: ok

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: if this is true second condition is also true, if this is false second condition is also false.

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: oops

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: ok

karantakalkar: if array is empty includes will give false, I can add condition for if array exists

karantakalkar:
@abhaychawla updated.

src/app/directives/has-permission/has-permission.directive.ts
abhaychawla: We can add a comment for description and remove the variable if it has no other purpose.

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,84 @@
...
+   * - User has special permission to access that feature.
+   * @returns {false}
+   * - Passed permission doesn't fall under either of above given permission grants.
+   * - Passed permission is null or undefined or incorrectly typed.
+   */
+  private hasPermission(permission: string) {
+    permission = permission.trim();
+    if (this.userPermissions && this.userPermissions.includes('ALL_FUNCTIONS')) {
+      return true;
+    } else if (permission !== null || permission !== undefined) {
```
abhaychawla, line 71:
```suggestion
    } else if (permission !== null && permission !== undefined) {
```

src/app/directives/has-permission/has-permission.directive.ts
abhaychawla: I meant in community app there are two conditions: 
hasPermission && !notPermissionFlag || !hasPermission && notPermissionFlag
The conditions are different and cannot be true at the same time. Also, can you check if its being actually used anywhere? If yes we can add a comment here explaining its usage to avoid confusion otherwise we will just remove it if its not being used.

abhaychawla:
Please resolve the conversations once you are done with the specific task in the comment.

src/app/directives/has-permission/has-permission.directive.ts
abhaychawla: Check for empty array is ok.

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: Yes, it's not being used, I felt like removing it but wasn't sure back then.

src/app/directives/has-permission/has-permission.directive.ts
karantakalkar: point, I think we should check only for null string then.

src/app/directives/has-permission/has-permission.directive.ts
```diff
@@ -0,0 +1,76 @@
...
+   * -`ALL_FUNCTIONS`: user is a Super user.
+   * -`ALL_FUNCTIONS_READ`: user has all read permissions and passed permission is 'read' type.
+   * - User has special permission to access that feature.
+   * @returns {false}
+   * - Passed permission doesn't fall under either of above given permission grants.
+   * - No value was passed to the has permission directive.
+   */
+  private hasPermission(permission: string) {
+    permission = permission.trim();
+    if (this.userPermissions && this.userPermissions.includes('ALL_FUNCTIONS')) {
```
abhaychawla, line 61:
```suggestion
    if (this.userPermissions.includes('ALL_FUNCTIONS')) {
```
As you said we won't need this as there will always be an empty array so this.userPermissions will return false.

karantakalkar: you want me to remove this.userPermissions?

abhaychawla: Yes

karantakalkar: ok

PR #1235, opened 2020-08-29
enhance: Replace select with angular material tree in CreateHoliday

zerefwayne:
#### Part 1: Convert the office list into a hierarchical object structure using office.hierarchy string.

I used a method similar to how a trie data structure is created to generate an object as specified in the documentation.

#### Part 2: Reading the Angular Material Checkbox Tree guidelines and sample code.

I went through the link given in description of the issue to understand how Angular Material Tree works.

#### Part 3: Integrating the specified code sections into CreateHoliday component and remove the previous select box.

I removed the select option and added all the required codes.

#### Part 4: Updating offices in HolidayForm object similar to what was being done before.

The format of checkListSelection Angular module is different from what we require, i.e. a simple list of strings (office ids). Therefore I create a function that updates the HolidayForm whenever checklist changes.

#1231 

## Screenshots, if any

The list generated by the Angular Material Tree comes in the console and I have tested it thoroughly.

image

## Checklist

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -32,30 +32,36 @@
             ],
             "styles": [
               "src/main.scss",
-              {"inject":false,
+              {
+                "inject": false,
                 "input": "src/theme/deeppurple-amber.scss",
                 "bundleName": "theme/deeppurple-amber"
-},
-              {"inject":false,
+              },
+              {
+                "inject": false,
                 "input": "src/theme/indigo-pink.scss",
                 "bundleName": "theme/indigo-pink"
-},
-              {"inject":false,
+              },
+              {
+                "inject": false,
                 "input": "src/theme/pink-bluegrey.scss",
                 "bundleName": "theme/pink-bluegrey"
-},
-              {"inject":false,
+              },
+              {
+                "inject": false,
                 "input": "src/theme/purple-green.scss",
                 "bundleName": "theme/purple-green"
-},
-              {"inject":false,
+              },
+              {
+                "inject": false,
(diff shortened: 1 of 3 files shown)


karantakalkar (changes requested):
Hi @zerefwayne good effort some changes in functionality are needed:

Node tree on dev instance https://dev.mifos.io/#/createholiday is as follows:
Screenshot from 2020-09-01 18-00-14

You need to check construction of first node, all peer nodes of Head Office (as in this screenshot) should be it's children.
Screenshot from 2020-09-01 18-00-00

In case you want to test it with different data change baseApiUrl in `environment.ts` to staging.mifos.io or demo.mifos.io (but login using dev.mifos.io only as auth on older servers is different). you can also check the respective community app (older interface)  functionality by visiting these urls.

zerefwayne:
@karantakalkar I somehow completely missed the root office condition. Thank you for pointing that out. I've fixed it, please have a look.

image

angular.json
```diff
@@ -32,30 +32,36 @@
             ],
```
karantakalkar, line 32: exclude this file from your commit

karantakalkar (changes requested):
You can test the API responses on community app instances, it may give off errors but at least request payload and error code should be the same for same input.

Currently we are submitting an array, which is giving of 500.

Screenshot from 2020-09-01 22-44-23


In community app API request there is an object array, so you need to map the offices array to become this:
Screenshot from 2020-09-01 22-44-11

karantakalkar:
Closing due to inactivity.

PR #1253, opened 2020-10-05
 Fixed view user page design

### LINKED ISSUE #1250: Fix view user page design
screenshot(1)

Abhirup-99:
## Description
Describe the changes made and why they were made instead of how they were made. List any dependencies that are required for this change.
UI updates.
#1250 

## Screenshots, if any
Screenshot from 2020-10-05 23-43-59

## Checklist

### DIFF
--- a/src/app/users/view-user/view-user.component.html
+++ b/src/app/users/view-user/view-user.component.html
@@ -45,16 +45,13 @@
           {{ userData.lastname }}
         </div>
 
-        <div *ngIf="userData.email">
 
-          <div fxFlex="50%" class="mat-body-strong">
-            Email
-          </div>
+        <div fxFlex="50%" class="mat-body-strong" *ngIf="userData.email">
+          Email
+        </div>
 
-          <div fxFlex="50%">
-            {{ userData.email }}
-          </div>
-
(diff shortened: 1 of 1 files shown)


src/app/users/view-user/view-user.component.html
```diff
@@ -45,16 +45,13 @@
...
-          <div fxFlex="50%">
-            {{ userData.email }}
-          </div>
+        <div fxFlex="50%" class="mat-body-strong" *ngIf="userData.email">
+          Email
+        </div>
 
+        <div fxFlex="50%" *ngIf="userData.email">
+          {{ userData.email }}
         </div>
```
Abhirup-99, line 55: @karantakalkar I have found removing this div, makes the entire thing working without any additional CSS changes. I agree this might not be the best method to achieve this, but in my opinion, the ngif is added for backward compatibility and while trying to create a user, I had to put an email as it was required. So, in this case, adding CSS to just maintain backward compatibility would be bad. So, I chose this approach. Correct me if I am wrong or if you don't like this approach, and I would rectify it accordingly.

karantakalkar (approved):

Issue #1229, opened 2020-08-27
Clients: create client address step options


karantakalkar:
In-congruency between data fetched on demo and staging instances causes address step (add) options to disappear on staging instance.

`demo.mifos.io`

![Screenshot from 2020-08-27 18-29-50](https://user-images.githubusercontent.com/54709463/91445395-ae47d080-e893-11ea-8767-d86b5e29f3e3.png)

`staging.mifos.io`

![Screenshot from 2020-08-27 18-30-25](https://user-images.githubusercontent.com/54709463/91445393-ad16a380-e893-11ea-94fe-6c9127a0c237.png)

karantakalkar:
@vorburger can the API response of this be made uniform across servers?

vorburger:
@karantakalkar I'm not involved in demo.mifos.io and staging.mifos.io, that's @edcable & Co. I only run https://www.fineract.dev, and invite you to use it because it's most up to date.

edcable:
@karantakalkar you should make the front end match the API for fineract.dev as the latest version of the back-end.

karantakalkar:
@vorburger 
demo.fineract.dev doesn't return address field, can you check?

![Screenshot from 2020-11-16 19-28-34](https://user-images.githubusercontent.com/54709463/99260955-10382680-2842-11eb-9c74-6da4ed4658a0.png)

 API request on https://dev.mifos.io/fineract-provider/api/v1/clients/template does return address fields.

vorburger:
@karantakalkar could you please email the Fineract Dev list about this? The code on fineract.dev is the latest, so perhaps this is some configuration option?

Issue #786, opened 2020-05-19
Support customizing baseApiUrl at run-time instead of build-time to be able to use different demo back-end servers


vorburger:
Following up on the discussion on #742 re. demo servers and https://wui.fineract.dev on https://www.fineract.dev : The reason why I did that and patched it like this https://github.com/openMF/web-app/compare/master...vorburger:fineract.dev is because https://openmf.github.io/web-app/, as per https://github.com/openMF/web-app#setting-up-a-local-server, is currently hard-coded to only https://demo.mifos.io in its https://github.com/openMF/web-app/blob/master/src/environments/environment.ts and to https://staging.mifos.io in its https://github.com/openMF/web-app/blob/master/src/environments/environment.prod.ts

This project would be very welcome to use https://www.fineract.dev instead, if you would like to change that hard-coded environment to be fineract.dev instead. The main advantage is that fineract.dev is automatically continuously updated, which could be useful as this projects adds new UIs for new back-end features, or even as we fix bugs there, as announced on https://markmail.org/message/7p32ufkp4wsq6cc5.

Another advantages of "driving" both end-users and developers to use fineract.dev is that, as they use it, errors are automatically reported, and I continuously create new bug reports for errors to be fixed in the back-end; see https://markmail.org/thread/6hkd2yvojngkeb6p and https://issues.apache.org/jira/browse/FINERACT-932.

An alternative could be to support customizing baseApiUrl at run-time instead of build-time to be able to use different demo back-end servers, and offer users an easy choice (without requiring rebuild).

vorburger:
On further thought, I suggest you really make it a baseApiUrl parameter (like community-app).. hardcoding makes it harder than it should be (because it requires a rebuild) for others to use this on their own servers.

edcable:
@abhaychawla @karantakalkar @Jov03 Can we prioritize completing this ticket? It would help make the forthcoming QA process before getting this released much easier.

karantakalkar:
@vorburger We are unable to login on `demo.mifos.io` and `staging.mifos.io` as auth has been changed to suit `dev.mifos.io`, I can add a feature to change baseApiUrl after login , in settings page.

Problem with using dev.mifos.io is lack of data, which means some features are inaccessible because there is no dummy case for it. If auth on demo and staging instances could be updated I can think of adding a dropdown/text for API on login page itself, just like language settings.

[...]

PR #1271, opened 2020-11-16
feat: add support for runtime server url changes

karantakalkar:
## Description

Pertaining to our discussion in #786 , we needed a way to change baseAPIUrl at runtime, this PR proposes to do the same, however instead of having server as a URL query param, it is a 'setting' instead, I think this approach would offer better UX.

Also considering the limitations of our routing structure, this seems viable. We can have a section on servers in the app, for users to be able to add/edit/remove their most frequently used domains.

#786 

## Screenshots, if any

Screenshot from 2020-11-16 22-22-41

## Checklist

### DIFF
--- a/src/app/web-app.component.ts
+++ b/src/app/web-app.component.ts
@@ -146,6 +146,16 @@ export class WebAppComponent implements OnInit {
     if (!localStorage.getItem('mifosXDateFormat')) {
       this.settingsService.setDateFormat('dd MMMM yyyy');
     }
+    if (!localStorage.getItem('mifosXServers')) {
+      this.settingsService.setServers([
+        'https://dev.mifos.io',
+        'https://demo.mifos.io',
+        'https://staging.mifos.io',
+        'https://mobile.mifos.io',
+        'https://demo.fineract.dev',
+        'https://localhost:8443'
+      ]);
+    }
   }
 
   logout() {
--- a/src/environments/.env.ts
+++ b/src/environments/.env.ts
@@ -1,3 +1,4 @@
 export default {
-  'npm_package_version': '0.0.0'
+  'mifos_x_version': '0.0.0',
+  'allow_switching_backend_instance': true
 };
--- a/src/environments/environment.ts
+++ b/src/environments/environment.ts
@@ -8,9 +8,10 @@ import env from './.env';
 
 export const environment = {
   production: false,
-  version: env.npm_package_version + '-dev',
+  version: env.mifos_x_version + '-dev',
   fineractPlatformTenantId: 'default',  // For connecting to server running elsewhere update the tenant identifier
-  baseApiUrl: 'https://dev.mifos.io',  // For connecting to server running elsewhere update the base API URL
+  baseApiUrl: JSON.parse(localStorage.getItem('mifosXServerURL')) || 'https://dev.mifos.io',  // For connecting to server running elsewhere update the base API URL
+  allowServerSwitch: env.allow_switching_backend_instance,
   apiProvider: '/fineract-provider/api',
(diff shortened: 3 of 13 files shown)


karantakalkar:
@vorburger I read the discussion on slack, I have a few hard-coded servers initialized in this, If you don't want the users to be able to add new servers then this PR is good to go, we can switch server on both login screen and settings page. 
Users can customize this list of servers only via rebuild.

karantakalkar:
@abhaychawla you may also want to see this.

src/app/web-app.component.ts
```diff
@@ -146,6 +146,16 @@ export class WebAppComponent implements OnInit {
...
+    if (!localStorage.getItem('mifosXServers')) {
+      this.settingsService.setServers([
+        'https://dev.mifos.io',
+        'https://demo.mifos.io',
+        'https://staging.mifos.io',
+        'https://mobile.mifos.io',
+        'https://demo.fineract.dev',
+        'https://localhost:8443'
+      ]);
+    }
```
karantakalkar, line 158: Here is the list of servers we are offering. Customizers can change this list to their own servers before build time.

src/environments/environment.ts
```diff
@@ -10,7 +10,8 @@ export const environment = {
   production: false,
   version: env.npm_package_version + '-dev',
   fineractPlatformTenantId: 'default',  // For connecting to server running elsewhere update the tenant identifier
-  baseApiUrl: 'https://dev.mifos.io',  // For connecting to server running elsewhere update the base API URL
+  baseApiUrl: JSON.parse(localStorage.getItem('mifosXServerURL')) || 'https://dev.mifos.io',  // For connecting to server running elsewhere update the base API URL
+  allowServerSwitch: true, // Set to false to disable changing baseApiUrl on runtime.
```
karantakalkar, line 14: Privilege to switch servers can be withdrawn by setting `allowServerSwitch` to false before build time.

Jov03:
@karantakalkar proceed as we discussed. move that configuration for allowing user to switch backend instance to env.ts and move all the configuration logic to login screen only.

karantakalkar:
If i move all logic to login screen...users wont be able to login into demo.mifos.io (because auth is different) which is our most data rich server.

Jov03:
See after login we cannot allow to switch backend instance because each instance would maintain there own user session. For eg. if lets say you logged in via demo.mifos.io then go to setting and change the backend server to demo.fineract.dev. Then demo.fineract.dev should consider you as an unauthorised user because you never logged into it, you logged into demo.mifos.io.

that's why we want this configuration to be available only at login page.

karantakalkar:
I agree, but that will hinder the testing process (unable to login to demo.mifos) maybe until the app gets tested we keep this makeshift ?

src/environments/.env.ts
```diff
@@ -1,3 +1,4 @@
 export default {
-  'npm_package_version': '0.0.0'
+  'mifos_x_version': '0.0.0',
+  'allow_switching_backend_instance': true
```
karantakalkar, line 3: Privilege to switch servers can be withdrawn by setting `allow_switching_backend_instance` to false before build time and specifying the server as `baseApiUrl`.

karantakalkar:
@Jov03 removed it from settings, I am a bit short on time to create a dialog that adds / displays and edits custom servers (opens on-click a gear icon beside mat-select) , that can come in later.

karantakalkar:
@edcable I have fixed #1229 in this, you can get address options on dev.mifos.io, however API response on fineract.dev, demo and staging is either incongruent or null for such options

Issue #1334, opened 2021-06-26
Collateral Module


BLasan:
**Is your feature request related to a problem? Please describe.**
It's better to have a collateral management module in order to manage the collateral data. This can be categorized in to 3 sub components.
1.) Product Level Collateral Module - Add, Edit, Delete and Update collateral

2.) Client Level Collateral Module - Add, Edit, Delete and Update client level collateral data which can then be used to attach to a particular loan.

3.) Loan Level Collateral Module - Attach the collateral in loan application creation process.

**Describe the solution you'd like**
See [PR](https://github.com/openMF/community-app/pull/3398)

**Describe alternatives you've considered**
Get a basic idea from the existing collateral management.

**Additional context**
Currently edit and delete options are not provided for loan level collateral once attached.

bharathcgowda:
completed

bharathcgowda:
@ramvr1256  @adamsaghy  thank you for working over this issue.

Here are the changes required to be made

1. Move the collateral selection screen at the loan level a step above the charges section( it is not related to charges, so we do not have to have it in the charge section)
2. There should not be any validations or configurations to mandate the collateral selection for a loan account
3. User will have the flexibility to select the collateral while creating the loan account from the collateral dropdown. if nothing is selected that means the loan doesn't have any collateral attached to it.
4. User should be able to create a loan account without selecting any collateral 
5. Only when the user attaches the collateral to the loan account, the collateral calculation should be done only for that loan account.

PR #1479, opened 2022-07-22
feat: collateral module

ramvr1256:
## Description
Collateral management module to manage the collateral data. This can be categorized into 3 sub-components.
1.) Product Level Collateral Module - Add, Edit, Delete, and Update collateral

2.) Client Level Collateral Module - Add, Edit, Delete, and Update client-level collateral data, which can then be used to attach to a particular loan.

3.) Loan Level Collateral Module - Attach the collateral in the loan application creation process.  Added validators to check whether loans are sufficient or not.

#1334 

## Screenshots, if any
Product Level
image

Overview of Collateral Products
image

Create Collateral Product
image

View Collateral Product
image

Edit Collateral Product
image

Delete Collateral Product
image

Client Level Collateral
image

Create Client collateral
image

Show all Clients collateral
image

View a client collateral
image

Edit Client Collateral
image

Delete Client Collateral
image

Loan Level Collateral
image
 The next button will remain disabled until you add collaterals with a value greater than your principal loan amount.

Adding collaterals to the loan
image

Validators for the max limit of collaterals a client has.
image

View Added collaterals
image

Remove added collaterals
image

Next button enabled after total collateral value exceeds the loan principal
image

## Checklist

### DIFF
--- a/src/app/products/collaterals/collaterals.component.html
+++ b/src/app/products/collaterals/collaterals.component.html
@@ -1,0 +1,56 @@
+<div class="container m-b-20" fxLayout="row" fxLayoutAlign="end" fxLayoutGap="20px">
+    <button mat-raised-button color="primary" [routerLink]="['create']" *mifosxHasPermission="'CREATE_CHARGE'">
+      <fa-icon icon="plus"></fa-icon>&nbsp;&nbsp;
+      Create Collateral
+    </button>
+  </div>
+  
+  <div class="container">
+  
+    <div fxLayout="row" fxLayoutGap="20px">
+      <mat-form-field fxFlex>
+        <mat-label>Filter</mat-label>
+        <input matInput (keyup)="applyFilter($event.target.value)">
+      </mat-form-field>
+    </div>
+  
+    <div class="mat-elevation-z8">
+  
+      <table mat-table [dataSource]="dataSource" matSort>
+  
+        <ng-container matColumnDef="name">
+          <th mat-header-cell *matHeaderCellDef mat-sort-header> Name </th>
+          <td mat-cell *matCellDef="let collateral"> {{ collateral.name }} </td>
+        </ng-container>
+  
+        <ng-container matColumnDef="quality">
+          <th mat-header-cell *matHeaderCellDef mat-sort-header> Quality/Type </th>
+          <td mat-cell *matCellDef="let collateral"> {{ collateral.quality }} </td>
+        </ng-container>
+
+        <ng-container matColumnDef="basePrice">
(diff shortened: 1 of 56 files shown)


bharathcgowda:
@karantakalkar @galovics @adamsaghy @josehernandezfintecheandomx could anyone of you help in the review of this PR?

adamsaghy:
@ramvr1256 I am sorry to not review earlier. Unfortunately it has a flaw... adding collateral became mandatory for all kind of loans without chechking whether the collateral module is enabled and it was configured to require guarantee

src/app/products/collaterals/collaterals.component.html
```diff
@@ -0,0 +1,56 @@
+<div class="container m-b-20" fxLayout="row" fxLayoutAlign="end" fxLayoutGap="20px">
+    <button mat-raised-button color="primary" [routerLink]="['create']" *mifosxHasPermission="'CREATE_CHARGE'">
```
adamsaghy, line 2: @ramvr1256 Wrong permission check

ramvr1256: I will correct this.

ramvr1256:
So, Should I remove the mandatory field validator from collateral ? Or how to know for which loan, the collateral module should be enabled?

adamsaghy:
@ramvr1256 I am not 100% familiar with the collateral module and concept, but i believe it should be optional and if there is no flag or anything to indicate whether it is required or not then that is a gap which should be corrected.

Would you mind if i put a checkbox on the Charges which says:
"Collateral is not required" and if it is checked the UI will let the user move forward?

adamsaghy:
My major problem is right now at backend side i dont see any validation or restriction with collaterals. So by postman i can create any kind of loan without providing any collateral, but now on the UI it is a requirement. However i am not really fond of solution where the UI is restricting something, but the backend not...

adamsaghy:
@ramvr1256 Would you consider to move Collateral to a new step on the Loan? I dont think it has anything to with "Charges". What do you think?

bharathcgowda:
@ramvr1256  @adamsaghy  thank you for working over this issue.

Here are the changes required to be made

1. Move the collateral selection screen at the loan level a step above the charges section( it is not related to charges, so we do not have to have it in the charge section)
2. There should not be any validations or configurations to mandate the collateral selection for a loan account
3. User will have the flexibility to select the collateral while creating the loan account from the collateral dropdown. if nothing is selected that means the loan doesn't have any collateral attached to it.
4. only when the user attaches the collateral to the loan account, the collateral calculation should be done only for that loan account.
5.

PR #1581, opened 2022-11-24
Fix the Change Password Functionality and add # as special character

IOhacker:
## Description
Fix the Change Password Functionality and add # as special character

#{Issue Number}

## Checklist

- [ ] If you have multiple commits please combine them into one commit by squashing them.

- [ ] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

### DIFF
--- a/src/app/self-service/users/user.service.ts
+++ b/src/app/self-service/users/user.service.ts
@@ -64,7 +64,7 @@ export class UserService {
    */
   changePassword(userId: string, passwordObj: any) {
 
-    return this.http.put(`/self/user/${userId}`, passwordObj);
+    return this.http.put(`/users/${userId}`, passwordObj);
   }
 
 }
--- a/src/app/shared/change-password-dialog/change-password-dialog.component.ts
+++ b/src/app/shared/change-password-dialog/change-password-dialog.component.ts
@@ -32,7 +32,7 @@ export class ChangePasswordDialogComponent implements OnInit {
   /** Change Password form */
   createChangePasswordForm() {
     this.changePasswordForm = this.formBuilder.group({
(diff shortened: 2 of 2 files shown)


src/app/self-service/users/user.service.ts
```diff
@@ -64,7 +64,7 @@ export class UserService {
    */
   changePassword(userId: string, passwordObj: any) {
 
-    return this.http.put(`/self/user/${userId}`, passwordObj);
+    return this.http.put(`/users/${userId}`, passwordObj);
```
adamsaghy, line 67: @IOhacker Are you sure we should call the /users API from self service? I think this is wrong. There is an Self service user API, which can be called:
`PUT /self/user`

IOhacker:
The self service is not implemented and I think it is for customers (like
web/mobile), not for staff users.

adamsaghy:
@IOhacker AFAIK it is implemented at backend: SelfUserApiResource.update()

Also there is at the UI /web-app/src/app/users/users.service.ts which has also change password functionality which calls the `  
changePassword(userId: string, passwordObj: any) {
    return this.http.put(`/users/${userId}`, passwordObj);
  }
 `

 That is for staff, no?
 
The self service API does not require the "userId" as it will be decided by the authenticated appuser.

Am is missing something here?

src/app/self-service/users/user.service.ts
IOhacker: @adamsaghy I think that if we are going to use the Self Service for changing the password of the User, then the we should use the Self Service Authentication for previously get the Access Token.

adamsaghy: I dont really know what the access token to do here... Behind the curtain  kind of the same happens, either API is called. My only concern here: If this is a UI for self service, I think the Self service API should be called.

There is UI view for (staff) user and for self-service. Kind of the same at backend side: there is API for self service and the regular API

adamsaghy: However, if you say this is not self-service and this is something really different, I am kind of okay with that as well. Like I said, kind of the same happens, either API is called

PR #1587, opened 2022-11-30
Client and External Id

josehernandezfintecheandomx:
## Description

In the client creation, we'll have the next scenarios for `externalId`:

- If the `externalId` value is added by the user this value will be used,
- If the `externalId` was not added by the user and the config to auto generate external Id (as UUID) is enabled, then we'll create this value as part of the client creation process, and as part of the response we'll have a new attribute for `resourceExternalId`

Screenshot 2022-11-30 at 22 56 40

- Long External Id value with an eye icon to display the full value
Screenshot 2022-12-01 at 22 36 27

- In the client view to show the full value If it's the case
Screenshot 2022-12-01 at 22 19 24

#{Issue Number}

## Checklist

- [ ] If you have multiple commits please combine them into one commit by squashing them.

- [ ] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

### DIFF
--- a/src/app/clients/client-stepper/client-general-step/client-general-step.component.html
+++ b/src/app/clients/client-stepper/client-general-step/client-general-step.component.html
@@ -16,7 +16,7 @@
 
     <mat-form-field fxFlex="48%">
       <mat-label>Legal Form</mat-label>
-      <mat-select formControlName="legalFormId">
+      <mat-select required formControlName="legalFormId">
         <mat-option *ngFor="let legalForm of legalFormOptions" [value]="legalForm.id">
           {{ legalForm.value }}
         </mat-option>
--- a/src/app/clients/client-stepper/client-general-step/client-general-step.component.ts
+++ b/src/app/clients/client-stepper/client-general-step/client-general-step.component.ts
@@ -1,6 +1,7 @@
 /** Angular Imports */
 import { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
 import { FormBuilder, FormGroup, Validators, FormControl } from '@angular/forms';
+import { ClientsService } from 'app/clients/clients.service';
 import { Dates } from 'app/core/utils/dates';
 
 /** Custom Services */
@@ -51,10 +52,12 @@ export class ClientGeneralStepComponent implements OnInit {
    * @param {FormBuilder} formBuilder Form Builder
    * @param {Dates} dateUtils Date Utils
    * @param {SettingsService} settingsService Setting service
+   * @param {ClientsService} clientService Client service
    */
   constructor(private formBuilder: FormBuilder,
               private dateUtils: Dates,
-              private settingsService: SettingsService) {
+              private settingsService: SettingsService,
+              private clientService: ClientsService) {
     this.setClientForm();
   }
 
@@ -71,7 +74,7 @@ export class ClientGeneralStepComponent implements OnInit {
     this.createClientForm = this.formBuilder.group({
       'officeId': ['', Validators.required],
       'staffId': [''],
-      'legalFormId': [''],
+      'legalFormId': ['', Validators.required],
       'isStaff': [false],
       'active': [false],
       'addSavings': [false],
@@ -142,6 +145,11 @@ export class ClientGeneralStepComponent implements OnInit {
       } else {
         this.createClientForm.removeControl('savingsProductId');
       }
+    });
+    this.createClientForm.get('officeId').valueChanges.subscribe((officeId: number) => {
+      this.clientService.getClientWithOfficeTemplate(officeId).subscribe((clientTemplate: any) => {
+        this.staffOptions = clientTemplate.staffOptions;
+      });
     });
   }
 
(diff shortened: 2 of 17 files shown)


src/app/clients/client-stepper/client-general-step/client-general-step.component.html
```diff
@@ -16,7 +16,7 @@
 
     <mat-form-field fxFlex="48%">
       <mat-label>Legal Form</mat-label>
-      <mat-select formControlName="legalFormId">
+      <mat-select required formControlName="legalFormId">
```
adamsaghy, line 19: Are you sure this is right? For non-entity client, the legal form cannot be required

src/app/clients/client-stepper/client-general-step/client-general-step.component.ts
```diff
@@ -71,7 +74,7 @@ export class ClientGeneralStepComponent implements OnInit {
     this.createClientForm = this.formBuilder.group({
       'officeId': ['', Validators.required],
       'staffId': [''],
-      'legalFormId': [''],
+      'legalFormId': ['', Validators.required],
```
adamsaghy, line 77: Are you sure this is right? For non-entity client, the legal form cannot be required

src/app/clients/clients-view/clients-view.component.html
```diff
@@ -29,20 +29,23 @@
...
           <p>
-            Client #:{{clientViewData.accountNo}} | External
-            Id: {{clientViewData.externalId}} | Staff:{{clientViewData.staffName || 'Unassigned'}}<br />
-            Activation Date : {{(clientViewData.activationDate) ? (clientViewData.activationDate | dateFormat) :'Not Activated'}}<br />
-            Member Of :
-            <span *ngFor="let group of clientViewData.groups">
-              <a>{{group.name}}</a>&nbsp;
+            <b>Office :</b> {{clientViewData.officeName}}<br />
+            <b>Client :</b> {{clientViewData.accountNo}}<br />
+            <b>External Id:</b> {{clientViewData.externalId | externalIdentifier}}<br />
```
adamsaghy, line 40: What is the value of the "externalIdentifier"?

src/app/clients/clients.service.ts
```diff
@@ -37,8 +37,12 @@ export class ClientsService {
     return this.http.get('/clients', { params: httpParams });
   }
 
-  getClientTemplate(): Observable<any> {
-    return this.http.get('/clients/template');
+  getClientTemplate(officeId: number): Observable<any> {
```
adamsaghy, line 40: It would be nicer if there would be 2 method. one with param and one without it, no?

src/app/pipes/external-identifier.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class ExternalIdentifierPipe implements PipeTransform {
+
+  transform(externalId: string): string {
+    if (!externalId) {
+      return '';
+    } else {
+      const regexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi;
+
+      if (regexExp.test(externalId)) {
```
adamsaghy, line 14: Is this necessary?

adamsaghy (changes requested):
See my comments

src/app/clients/client-stepper/client-general-step/client-general-step.component.html
josehernandezfintecheandomx: In both cases (Person and Entity) the Legal Form is required

src/app/clients/client-stepper/client-general-step/client-general-step.component.ts
josehernandezfintecheandomx: In both cases (Person and Entity) the Legal Form is required

src/app/clients/clients-view/clients-view.component.html
josehernandezfintecheandomx: The `externalIdentifier` is a new `Pipe` mainly to format (or cut) the long UUID auto generated and just present the latest part.
- If the externalId value is null then will return null, 
- If the externalId value has not an UUID format, then return the same value, and
- If the externalId value has an UUID format, then It is spliced and  return the latest part

And It is a `Pipe` because the externalId value is used in different parts of the UI

src/app/pipes/external-identifier.pipe.ts
adamsaghy: Do we really need to truncate the external id?

src/app/clients/clients.service.ts
josehernandezfintecheandomx: Done! Updated

src/app/pipes/external-identifier.pipe.ts
josehernandezfintecheandomx: Yes, the Idea is to detect an externalId value with an UUID format and split this to get the latest part of the long value

adamsaghy: okay, but why? Dont we have enough space to show it? I dont think it will be straightforward, it is not the externalId, just the last part of it... no?

josehernandezfintecheandomx: Screenshot 2022-12-01 at 22 36 27

adamsaghy (approved):
LGTM

Issue #1649, opened 2023-01-21
Apache Superset Support for Dashboards and Insights


IOhacker:
**Is your feature request related to a problem? Please describe.**
Apache Superset is a great visualization tools which has a easy setup and an active commmunity so then the request is to support it for creating Dashboards and Insights using the Data generated by Apache Fineract. It will help to display information for decisioning about loans, savings, users (demographic data) and non functional actions done in Apache fineract (i.e. audit trail logging). 

**Describe the solution you'd like**
Using the Embedded SDK will help to acomplish this feature request https://www.npmjs.com/package/@superset-ui/embedded-sdk

**Describe alternatives you've considered**
Pentaho CDE
Redash

**Additional context**
NA

edcable:
Re-opening issue as @PC-11-00 is taking up work on this as part of 2023 Google Summer of Code

adamsaghy:
@edcable @IOhacker I have seen this PR: https://github.com/openMF/web-app/pull/1849, but i am having concerns whether it is a good idea.

The implementation here was all the authentication details to be stored at UI side which is serious security concern...

In the other hand are we really sure the web-app should support 2 pretty separate app:
- fineract-backend
- apache superset backend?

PR #1849, opened 2023-08-24
feat: added apache superset support for dashboard and insights

PC-11-00:
## Description
now we have a great visualization tool, it will help to dislpay information for decisioning about loans, savings, users.

#1649 

## Screenshots, if any

https://github.com/openMF/web-app/assets/76156941/dc469481-6320-4517-b5c0-8c18f9127dd0

## Checklist

### DIFF
--- a/src/app/core/authentication/authentication.interceptor.ts
+++ b/src/app/core/authentication/authentication.interceptor.ts
@@ -27,7 +27,7 @@ const twoFactorAccessTokenHeader = 'Fineract-Platform-TFA-Token';
 @Injectable()
 export class AuthenticationInterceptor implements HttpInterceptor {
 
-  constructor(private settingsService: SettingsService) {}
+  constructor(private settingsService: SettingsService) { }
 
   /**
    * Intercepts a Http request and sets the request headers.
@@ -36,7 +36,14 @@ export class AuthenticationInterceptor implements HttpInterceptor {
     if (this.settingsService.tenantIdentifier) {
       httpOptions.headers['Fineract-Platform-TenantId'] = this.settingsService.tenantIdentifier;
     }
-    request = request.clone({ setHeaders: httpOptions.headers });
+
+    /**
+     * This is made because we added apache superset and in superset we needed to send request to superset backend(localhost:8088) 
+     */
+    if (request.url.includes("fineract-provider")) {
+      request = request.clone({ setHeaders: httpOptions.headers });
+    }
+
     return next.handle(request);
   }
 
--- a/src/app/home/dashboard/dashboard.component.html
+++ b/src/app/home/dashboard/dashboard.component.html
@@ -8,12 +8,13 @@
 
         <mat-card-header class="header">
           <h3><fa-icon icon="bolt"></fa-icon>
-          &nbsp;&nbsp; Recent Activities</h3>
+            &nbsp;&nbsp; Recent Activities</h3>
         </mat-card-header>
 
         <mat-card-content>
           <mat-action-list class="list">
-            <button mat-list-item *ngFor="let activity of recentActivities" (click)="navigatetoActivity(activity)"> {{ activity | urlToString }} </button>
+            <button mat-list-item *ngFor="let activity of recentActivities" (click)="navigatetoActivity(activity)"> {{
+              activity | urlToString }} </button>
           </mat-action-list>
         </mat-card-content>
 
@@ -23,16 +24,16 @@
 
         <mat-card-header class="header">
           <h3><fa-icon icon="laptop"></fa-icon>
-          &nbsp;&nbsp; Frequent Activities</h3>
+            &nbsp;&nbsp; Frequent Activities</h3>
         </mat-card-header>
 
         <mat-card-content>
           <mat-action-list class="list">
-            <button mat-list-item *ngFor="let activity of frequentActivities"
-            [routerLink]="activity"> {{ activity | urlToString }} </button>
+            <button mat-list-item *ngFor="let activity of frequentActivities" [routerLink]="activity"> {{ activity |
+              urlToString }} </button>
           </mat-action-list>
         </mat-card-content>
-        
+
       </mat-card>
 
       <mifosx-amount-disbursed-pie></mifosx-amount-disbursed-pie>
@@ -46,7 +47,7 @@
       <mifosx-amount-collected-pie></mifosx-amount-collected-pie>
 
     </div>
-    
+    <div *ngIf="supersetEnable"><mifosx-superset></mifosx-superset></div>
   </mat-card>
 
 </div>
--- a/src/app/home/dashboard/dashboard.component.html
+++ b/src/app/home/dashboard/dashboard.component.html
@@ -47,7 +47,7 @@
       <mifosx-amount-collected-pie></mifosx-amount-collected-pie>
 
     </div>
-    <div *ngIf="supersetEnable"><mifosx-superset></mifosx-superset></div>
+    <div *ngIf="!supersetEnable"><mifosx-superset></mifosx-superset></div>
   </mat-card>
 
 </div>
--- a/src/app/home/superset/superset.service.ts
+++ b/src/app/home/superset/superset.service.ts
@@ -1,0 +1,119 @@
+/** Angular Imports */
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpHeaders } from '@angular/common/http';
+
+/** rxjs Imports */
+import { Observable, throwError } from 'rxjs';
+import { catchError, switchMap } from 'rxjs/operators';
+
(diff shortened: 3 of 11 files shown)


src/app/core/authentication/authentication.interceptor.ts
```diff
@@ -36,7 +36,14 @@ export class AuthenticationInterceptor implements HttpInterceptor {
     if (this.settingsService.tenantIdentifier) {
       httpOptions.headers['Fineract-Platform-TenantId'] = this.settingsService.tenantIdentifier;
     }
-    request = request.clone({ setHeaders: httpOptions.headers });
+
+    /**
+     * This is made because we added apache superset and in superset we needed to send request to superset backend(localhost:8088) 
+     */
+    if (request.url.includes("fineract-provider")) {
```
adamsaghy, line 43: i dont see moving the `request = request.clone({ setHeaders: httpOptions.headers });` behind a condition what to do with apache superset...

src/app/home/dashboard/dashboard.component.html
```diff
@@ -46,7 +47,7 @@ <h3><fa-icon icon="laptop"></fa-icon>
       <mifosx-amount-collected-pie></mifosx-amount-collected-pie>
 
     </div>
-    
+    <div *ngIf="!supersetEnable"><mifosx-superset></mifosx-superset></div>
```
adamsaghy, line 50: This seems wrong to me. The template to be shown if the condition is true...no?

src/app/home/superset/superset.service.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class SupersetService {
+  
+  /**
+   * API URL of Superset to send request
+   */
+  private apiUrl = 'http://localhost:8088/api/v1/security';
```
adamsaghy, line 23: Please move all of the configuration into a property file and do not hardcode it.

src/app/home/superset/superset.service.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  constructor(private http: HttpClient) { }
+
+  /**
+   * 
+   * @returns { access token }
+   */
+  private fetchAccessToken(): Observable<any> {
+    const body = {
+      "username": "admin",
```
adamsaghy, line 36: I am not sure it is a good idea to store credential information at UI side... anybody can read them and abuse...

src/environments/environment.ts
```diff
@@ -30,6 +30,10 @@ export const environment = {
   supportedLanguages: window['env']['supportedLanguages'] || 'en-US,es-MX,fr-FR',
   preloadClients: window['env']['preloadClients'] || true,
 
+  superset: {
+    enabled: window['env']['supersetEnabled'] || true
```
adamsaghy, line 34: Please have this functionality `disabled` by default. This seems to me an optional feature.

adamsaghy (changes requested):
Please kindly review my questions and concerns!

To be honest, i am not sure this functionality to be embedded into fineract... i am a little worried about the security (credentials to be stored at client side) and it is really easy to abuse this...

src/app/core/authentication/authentication.interceptor.ts
PC-11-00: Actually, apache superset will run as an independent app so we didn't want to send "fineract-provider" as a header because it will not work otherwise. for embedding the apache superset we are not sending request to the fineract ,we are running the apache superset at the port :8088 and then sending the request to it, that's why added this

src/app/home/dashboard/dashboard.component.html
PC-11-00: Actually, we just need to remove ! to show the superset and @josehernandezfintecheandomx suggested me, keep it as an environment variable ,but true and false were not working, so I kept it as true in the environment variable and added: !

PC-11-00:
Actually, it will not embed into fineract backend ,but  this will run as an independent app

src/app/home/superset/superset.service.ts
PC-11-00: Do you want me to make config.json and move the URL, username, and password into it?

src/environments/environment.ts
PC-11-00: It is disabled in the HTML file by using:! sign, so it will not shown by default

PC-11-00:
@josehernandezfintecheandomx knows everything about this PR

adamsaghy:
> Actually, it will not embed into fineract backend ,but this will run as an independent app

I was referring to the UI (web-app).

src/app/home/dashboard/dashboard.component.html
adamsaghy: I am not sure i am following you, why the true or false would not work for ngIf which is working based on the variable value is true of false...

src/app/home/superset/superset.service.ts
adamsaghy: That would be better, but still... It is on UI...anybody can retrieve these information... this is not secure... i cant imagine anybody would want to share with the world the username and password....

src/environments/environment.ts
adamsaghy: It is still confusing... maybe you should name it as disabled then or something... but you understand where i am coming from right?

Here is a configuration which value is `true`... superset.enabled = true and when it is true, then the superset is disabled... thats...wrong :D

src/app/home/dashboard/dashboard.component.html
PC-11-00: I know theoretically this should work, but me and @josehernandezfintecheandomx both tried this but it was not working

src/app/home/superset/superset.service.ts
PC-11-00: Hmm then we can make a service in our fineract backend to access this URL, username, and password in the frontend

PC-11-00: Becuase we need to send the request to some other app

adamsaghy: If you want my honest opinion: this feature is not really related to fineract. I would not try to combine them together... i might be wrong, but i dont see what is the benefit here... also hence the security concerns, noone would use it in production...or publicly deployed...

PC-11-00: okay then I will add a config.json file

adamsaghy: @bharathcgowda @edcable @josehernandezfintecheandomx What are your opinions about this?

bharathcgowda: @adamsaghy @PC-11-00   I do not have a solution to suggest by but I feel it is not good to store username and password anywhere in the code or in config.json file.
It is always gets blocks during security screening.

adamsaghy: Yeah. that is the main reason i dont really think it is a feature to be added... :/ This looks to me more like a separate application...

PC-11-00: @adamsaghy Yeah that's correct apache superset is a different application, which  I am embedding into the web app , it's not a feature, to run Apache superset dashboard, we need to run Apache superset and web app separately 
You can read this blog: https://medium.com/@chaudharypushpendra.11.10.2000/embedding-of-apache-superset-dashboard-in-the-mifos-initiative-angular-web-app-b9259f1f1f1b

PC-11-00: You can check their git repo: https://github.com/apache/superset
And the issue: https://github.com/openMF/web-app/issues/1649 was to embed the apache superset

adamsaghy: so if i understand correctly, adding this dashboard into the web-app was your Gsoc assignment? Can you please share more details about what was exactly the job here? 
Nevertheless hardcoding authentication details into UI (even if it is config.json) file is wrong, and a huge security risk!

I dont really see why should the wep-app serve 2 total different application:
- Fineract backend
- Apache superset backend

adamsaghy: Is there any way to remove the credentials from the UI? i dont mind if the user must log into manually into the apache superset or the authentication is sync within the superset and fineract, but i just cant wrap my head around such an idea to hardcode into UI any of these sensitive informations...

PC-11-00: I know it is wrong to hardcode the username and password, but I asked @josehernandezfintecheandomx to make a service in fineract backend from where we can fetch the Guest Token of Apache Superset, but he denied making that
but we need to store those credentials  somewhere and  I didn't find any solution and so I hard coded.

adamsaghy: I have dropped a message to Ed and Victor regarding my concerns with this feature to be added to the UI

PR #1887, opened 2023-10-30
Added configurable warning message

Vectorquantiy:
## Description
Implemented a configurable legal warning popup that appears after the login screen. The content of the popup message is configurable and customizable, allowing for easy updates and modifications of the information displayed.

### DIFF
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -21,6 +21,11 @@ export const environment = {
     enabled: false,  // For connecting to Mifos X using OAuth2 Authentication change the value to true
     serverUrl: ''
   },
+  warningDialog: {
+    title: 'Warning',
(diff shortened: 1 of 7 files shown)


src/environments/environment.prod.ts
```diff
@@ -21,6 +21,11 @@ export const environment = {
     enabled: false,  // For connecting to Mifos X using OAuth2 Authentication change the value to true
     serverUrl: ''
   },
+  warningDialog: {
```
adamsaghy, line 24: Could you please provide a default message which is not a placeholder?

**Example:**
This system is for authorized use only. Unauthorized access will result in possible legal action. By accessing this system, you acknowledge that 
you are authorized to do so and that all data stored and processed here is confidential.

adamsaghy (changes requested):
Please kindly see my comment!

src/environments/environment.prod.ts
Vectorquantiy: I added the requested message.

adamsaghy (approved):

Issue #1895, opened 2023-11-13
Support JSON object on UI


peter-santa-dpc:
**Background**
It is now supported by Fineract, to store JSON typed object in a Postgres DB field of a data table

- https://github.com/apache/fineract/pull/3513 
- https://issues.apache.org/jira/browse/FINERACT-1996

Currently the UI shows [object Object] instead of the data.

**Goal**
The UI should show a formatted JSON.

PC-11-00:
Closing as it is completed

PR #1902, opened 2023-11-17
datatable-single-row now supports json object

Vectorquantiy:
## Description
datatable-single-row.component.html now shows json obejcts correctly.

#1895


adamsaghy (approved):
LGTM

Issue #1901, opened 2023-11-16
UI Generates outside link in some circumstances


peter-santa-dpc:
## Description
We have a data table, that is assigned to a savings transaction - but **the issue may be valid for all data table views**.
There is a string/text field, in which there is a string, e.g.: "TXNB:HU10011111". The UI makes it as link.

## Steps to Reproduce

1. Have a data table, assigned to a savings transaction
2. Have a field in the data table, that can store String. E.g.:
{
            "name": "special_identifier",
            "type": "String",
            "mandatory": false,
            "length": 2048,
            "indexed": true,
            "unique":false
        }
3. Insert a transaction, and insert a record in the data table. Let the value of special_identifier be "TXNB:HU10011111".

## Expected Behaviour
The string is shown on the UI, as it is stored.

## Actual Behaviour
A link is generated to the value. If you click on it, it tries to open the "link".

## Settings
- Mifos X version: c8703adbc156a1bbc4f0d06dbd5250e37934aaac
- Browser used: Firefox 119.0.1 (64-bit)

## Screenshots, if any
![image](https://github.com/openMF/web-app/assets/45198877/6af1e2ca-25fb-4698-a107-db9ffcf74172)

--------------------

![image-20231115-110515](https://github.com/openMF/web-app/assets/45198877/f497568c-5c97-4191-80d6-e3eca4e79e7d)

PC-11-00:
The issue is resolved.

PR #1903, opened 2023-11-17
url validation fixed

Vectorquantiy:
## Description
Had to change url validation because when the input string had ":" in it, it was enough to return true, resulting in an invalid url.

#1901


adamsaghy (approved):
LGTM

Issue #1899, opened 2023-11-15
Savings Transaction - assigned data table - ID field shown as decimal


peter-santa-dpc:
## Description
Data tables can be assigned to savings transaction.
There is a view for data table details. The ID of the "parent" transaction is shown on the view. It is shown with 2 decimals, instead of integer.

## Steps to Reproduce

1. Have a Savings Account
2. Have data table assigned to savings accounts
3. Have a transaction inserted, have an entry inserted in the related data table.

## Expected Behaviour
The ID of the savings transaction is shown without decimals.

## Actual Behaviour
The ID of the savings transaction is shown with decimals.

## Settings
- Mifos X version: c8703adbc156a1bbc4f0d06dbd5250e37934aaac
- Browser used: Firefox 119.0.1 (64-bit)

## Screenshots, if any
![image](https://github.com/openMF/web-app/assets/45198877/979dbaa4-ee15-44b6-a610-da38d85a18d7)

peter-santa-dpc:
https://issues.apache.org/jira/browse/FINERACT-2014 should resolve this issue.

PC-11-00:
Closing as it is resolved.

PR #1908, opened 2023-11-24
Id field shown as decimal

Vectorquantiy:
## Description
Separated integer and decimal formatting on datatable-single-row component.

#1899


adamsaghy (approved):
LGTM

PR #2029, opened 2024-03-10
[Feat]: Migrate the web-app from angular-14 to angular-15

vikashsprem:
## Description
Migrate the web-app from angular-14 to angular-15


PC-11-00:
Why are you changing Angular version?

PR #2094, opened 2024-05-21
Production login uses same domain. Fixes #2093

### LINKED ISSUE #2093: Login fails on production server setup (non-Docker)
## Description
On multi-tenant production setup, where each tenant runs on a sub-domain of the same parent, e.g abc.mifos.io xyz.mifos.io, we want each tenant to be insulated from the others, so no option to switch backend server. We also want to connect to Fineract on the same sub-domain by default https://xyz.mifos.io connects to https://xyz.mifos.io/fineract-provider/api/v1.

In such a VM setup with Nginx (non-Docker), login tries to use https://demo.fineract.dev/fineract-provider/api/v1/authentication by default and fails, because this URL is down. It should connect to the current production domain by default. 

## Steps to Reproduce
Setup the web-app on a Server/VM with Nginx as a multi-tenant on subdomain setup (using prefix tenantId.domain.com) with Fineract also on the same domain (non-Docker). Since this is for production, would like each tenant to not need to select the backend but automatically to the same (sub)domain Fineract platform.

## Expected Behaviour
Web-app login should connect to Fineract backend server on the same domain and authenticate successfully.

## Actual Behaviour
Web-app attempts to connect to https://demo.fineract.dev which has been down since a few years and fails to login

## Settings
- Mifos X version: 23.12.31
- Browser used: Chrome

## Screenshots, if any

terencemo:
## Description
Since for production in a multi-tenant setup where each tenant wants to connect to the same tenant backend, we want to connect to the same domain backend and don't want the option to select the backend. So allowServerSwitch is set to false and default backend is set to 'https://' + window.location.hostname.

#2093 

## Checklist

### DIFF
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -12,8 +12,8 @@ export const environment = {
   baseApiUrls: window['env']['fineractApiUrls'] ||
   'https://dev.mifos.io,https://demo.mifos.io,https://qa.mifos.io,https://staging.mifos.io,https://mobile.mifos.io,https://demo.fineract.dev,https://localhost:8443',
   // For connecting to server running elsewhere set the base API URL
-  baseApiUrl: window['env']['fineractApiUrl'] || 'https://demo.fineract.dev',
-  allowServerSwitch: env.allow_switching_backend_instance,
+  baseApiUrl: window['env']['fineractApiUrl'] || 'https://' + window.location.hostname,
+  allowServerSwitch: false,
   apiProvider: window['env']['apiProvider'] || '/fineract-provider/api',
   apiVersion: window['env']['apiVersion'] || '/v1',
   serverUrl: '',


terencemo:
@alberto-art3ch kindly review

alberto-art3ch:
> @alberto-art3ch @adamsaghy kindly review

This is possible using an environment variable to be used in the dropdown Tenant Id options:

1. Set the right value for the environment variables `FINERACT_PLATFORM_TENANT_IDENTIFIER` and `FINERACT_PLATFORM_TENANTS_IDENTIFIER`  as follow

`FINERACT_PLATFORM_TENANTS_IDENTIFIER` Is the Tenant Id list sparated by coma, ie: 
    ` FINERACT_PLATFORM_TENANTS_IDENTIFIER=tenant1,tenant2,tenant3,tenant4`
     
`FINERACT_PLATFORM_TENANT_IDENTIFIER` is the Tenant Id used by default, in this case must be one of the tenant Id in the list, ie:
     `FINERACT_PLATFORM_TENANT_IDENTIFIER=tenant2`
     
You must to restart the ng service or whatever service that delivers the web-app application

Finally you will be able to see an enabled and populated dropdown Tenant Id menu like the next image

Screenshot 2024-05-21 at 1 56 35 p m

alberto-art3ch:
Hello Terence!

Please review the PDF attached and any comment or doubt please let me know

Best Regards
Alberto

terenceamb:
> Setting the variables must be done in the env.js file not in this file. You are replacing a variable with a constant.
> 
> FROM allowServerSwitch: env.allow_switching_backend_instance, TO allowServerSwitch: false
> 
> which is not correct.

Thanks @IOhacker  for all your good work and contribution to Mifos. In this PR, Dev still has backend selection with dropdown by default and the variable remains in environment.ts (only environment.prod.ts is proposed to be changed). It would certainly be a good feature to be able to select from different backends during development or for demo purposes. For Prod my proposal was to set allowServerSwitch to false - this changes the variable to a constant, which you say is not correct. My counter-question is do we really want Production users to have a choice of selecting a different backend by default?

Let's say a Fintech Company Xyz had a multi-tenant setup with 10 tenants. One of them is Abc Bank. Would Xyz Fintech like to provide Abc Bank with a simple direct login where Abc Bank staff login leads to Abc Bank backend? Or would Xyz Fintech solution for Abc Bank include a dropdown to Abc Bank employees of Abc Bank plus the other 9 Financial Organizations? As a Fintech firm involved with Mifos for about 15 years, I'd say I would like to provide each Bank with a simple login screen without a backend selection dropdown. @edcable can we get feedback from other Fintech companies and/or Banks/FIs on whether their production setup login screens should have backend selection dropdown or just a simple screen with credentials and language selection only?

Now @IOhacker maybe you're suggesting editing env.js but env.js is generated by version.js and the first line says this: IMPORTANT: THIS FILE IS AUTO GENERATED! DO NOT MANUALLY EDIT OR CHECKIN!
And this env.js comes with the option env.allow_switching_backend_instance = true
So do you rather suggest we modify version.js to generate env.js with env.allow_switching_backend_instance = false?
For a Production build, this makes sense to me, but for Dev, I think its good to have the option to select the backend.

terencemo:
> Setting the variables must be done in the env.js file not in this file. You are replacing a variable with a constant.
> 
> FROM allowServerSwitch: env.allow_switching_backend_instance, TO allowServerSwitch: false
> 
> which is not correct.

Thanks @IOhacker for all your good work and contribution to Mifos. In this PR, Dev still has backend selection with dropdown by default and the variable remains in environment.ts (only environment.prod.ts is proposed to be changed). It would certainly be a good feature to be able to select from different backends during development or for demo purposes. For Prod my proposal was to set allowServerSwitch to false - this changes the variable to a constant, which you say is not correct. My counter-question is do we really want Production users to have a choice of selecting a different backend by default?

Let's say a Fintech Company Xyz had a multi-tenant setup with 10 tenants. One of them is Abc Bank. Would Xyz Fintech like to provide Abc Bank with a simple direct login where Abc Bank staff login leads to Abc Bank backend? Or would Xyz Fintech solution for Abc Bank include a dropdown to Abc Bank employees of Abc Bank plus the other 9 Financial Organizations? As a Fintech firm involved with Mifos for about 15 years, I'd say I would like to provide each Bank with a simple login screen without a backend selection dropdown. @edcable can we get feedback from other Fintech companies and/or Banks/FIs on whether their production setup login screens should have backend selection dropdown or just a simple screen with credentials and language selection only?

Now @IOhacker maybe you're suggesting editing env.js but env.js is generated by version.js and the first line says this: IMPORTANT: THIS FILE IS AUTO GENERATED! DO NOT MANUALLY EDIT OR CHECKIN!
And this env.js comes with the option env.allow_switching_backend_instance = true
So do you rather suggest we modify version.js to generate env.js with env.allow_switching_backend_instance = false?
For a Production build, this makes sense to me, but for Dev, I think its good to have the option to select the backend.

PR #2182, opened 2024-10-23
update for Docker Compose v2

meonkeys:
using the new recommended command-line syntax since v1 is EOL.

`docker-compose` runs the old Python v1 code.

v2 is the new go-based, built-in orchestration tool.

`docker compose` (no dash) makes sure v2 is used.

See also:

* https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/
* https://docs.docker.com/compose/releases/migrate/


IOhacker (approved):
LGTM

PR #2203, opened 2024-11-26
FINERACT-1981: Introduce Interest should not be calculated on past due principal amount on API

janez89:
This P.R. introduces a new flag under the interest recalculation section on the loan product.

It also contains some minor fixes.


IOhacker:
I think that language translation can be implemented using google translate to avoid any technical debt

PR #2217, opened 2024-12-14
fix: allowPartialPeriodInterestCalculation

rhopman:
## Description
Due to a typo, the allowPartialPeriodInterestCalculation field was not processed correctly.

## Checklist


adamsaghy:
@rhopman Kindly rebase this PR.

adamsaghy (changes requested):
This fix requires changes at the backend side as well, as this field has the typo there as well!

rhopman:
@adamsaghy good catch. I thought I checked for that but apparently I made a mistake. Let me look into this. The backend change would be breaking, so that would not be an easy fix.

rhopman:
@adamsaghy I did a bit more research. Although the typo is indeed also present in the Fineract backend, it doesn't seem to have leaked into the API. I checked the following endpoints:

- `GET /v1/loans/template`
- `GET /v1/loans/{loanId}`
- `GET /v1/loanproducts/template`
- `GET /v1/loanproducts/{productId}`

and they all use `allowPartialPeriodInterestCalculation` (so without the typo).

IOhacker:
@adamsaghy any other comment about the response sent by @rhopman

adamsaghy:
@IOhacker @rhopman I am so sorry, but i havent had time to review this ticket... i see some API has the typo and some does not have it... so i reckon the fix should be something like if we wanna make sure backward compatibility :
- send the typo and the fixed field name as well
- accept the typo and the fixed field name as well

However, this would not be the 1st case when we kind of decide to break it and just introduce the change on the Fineract DEV mail list.

@rhopman @IOhacker What do you think? Which would be the best solution?

rhopman:
>     * send the typo and the fixed field name as well

Unfortunately, adding the fixed field name to the payloads causes Fineract to give the following error: `The parameter allowPartialPeriodInterestCalculation is not supported.`
 
>     * accept the typo and the fixed field name as well

The `GET` endpoints do not have the typo so I don't think there is any need to accept it.

@adamsaghy did you review the code I pushed on December 19th? I think this is the best solution for now. It fixes the typo where possible, and where needed it puts it back in right before sending the payload. Ideally, we will make this workaround dependent on the backend version once we know in which version it will be fixed.

What do you think?

adamsaghy (approved):
LGTM

PR #2255, opened 2025-01-29
build: remove tailwindcss dependencies

rhopman:
## Description

Removed direct dependencies to `tailwindcss` and `@tailwindcss/forms`, since they don't seem to be used.

I don't see references to tailwind anywhere. I tested manually after removing the dependencies, and everything seems to look exactly the same. So I think these dependencies can be safely removed.

@josehernandezfintecheandomx / @alberto-art3ch I think you introduced these dependencies in 2023, so perhaps you can confirm?

## Checklist


NeftaliYagua:
In my opinion, I would like the next major version to be designed with TailwindCSS, that would be great, although everything would have to be rewritten, in my opinion Angular Material is a bit square. :)

PR #2266, opened 2025-02-14
fix: Javascript "is undefined" error

steinwinde:
Make Chart of Accounts Javascript run without error when no accounts exist

Fixes: WEB-10


PC-11-00:
@steinwinde one PR = one commit
Please squash the commits

PR #2267, opened 2025-02-16
fix: some issues with the Configuration Wizard

steinwinde:
Two wizards just ceased showing popups. Another wizard ended in a 404. Many of the popups had problems with the labels (e.g. literally showed the label instead of the translation) or were not correctly positioned, often due to Javascript errors. This fixes all major problems with the wizards. Remaining issues will be subject to at least one separate Jira.

Fixes: WEB-11


PC-11-00 (approved):
LGTM!

PR #2275, opened 2025-02-23
fix: login form

pranav-deshmukh:
## Description

View password option works correctly. 
Login form  submits only on clicking Login
Minimum 8 charcters limit (which was already set for passwords) is now mentioned to users.

## Fixes issue: WEB-34
Before fix:

https://github.com/user-attachments/assets/8fb9373c-f357-4b50-95c1-82710d6c1a17

After fix:
https://github.com/user-attachments/assets/a82765e9-9a50-491f-ba74-03468fb25f75

### DIFF
--- a/src/assets/translations/en-US.json
+++ b/src/assets/translations/en-US.json
@@ -772,6 +772,7 @@
       "is": "is",
       "Is": "is",
       "is required": "is required",
+      "pass is required": "is required(min length 8)",
       "least one uppercase": "least one uppercase",
       "lowercase and special character": "lowercase and special character",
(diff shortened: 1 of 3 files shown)


src/assets/translations/en-US.json
```diff
@@ -772,6 +772,7 @@
       "is": "is",
       "Is": "is",
       "is required": "is required",
+      "pass is required": "is required(min length 8)",
```
oluexpert99, line 775: can we add for other languages?

alberto-art3ch:
LGTM

PR #2301, opened 2025-03-06
fix:en-US translations for missing event types

Ashmn07:
## Description

Event types were missing in the translations file, so the missing values have been added just to the en-US file

Fixes issue WEB-16

## Screenshots
Before Fix
image

After fix
image


IOhacker:
LGTM

PR #2323, opened 2025-03-16
fix: Add possibility to create GLIM account applications

steinwinde:
This is a first step to support GLIM account applications. It enables a user to create a GLIM account application and provides a starting point for a discussion of the feature, which has not even been working for the community app and is insufficiently documented. This is part of a larger effort to improve group account management.

Partially addresses: WEB-9


IOhacker:
@steinwinde this happen when you try to create a new loan (an usual loan application)

PR #2340, opened 2025-03-22
fix: Prevent GLIM and GSIM without members

steinwinde:
Makes sure GLIM and GSIM application menu item doesn't appear without members. (This has already been in place for JLG.) Make sure GSIM can't be saved without selected members.

Fixes: WEB-62


donmbelembe:
Hello @steinwinde , @IOhacker  This commit actually introduced a bug, when trying to add a product to an existing client there is an error and can't proceed to the next step.

image

PR #2389, opened 2025-04-07
fix: Improve floating rate creation and editing UI

steinwinde:
Better datepicker value; disabled submit without period; removed minDate in main form.

Partially addresses: WEB-103


steinwinde:
Noticed a mistake reviewing.

PR #2390, opened 2025-04-07
fix: Improve floating rate creation and editing UI

steinwinde:
Better datepicker value; disabled submit without period; removed minDate in main form.

Partially addresses: WEB-103


steinwinde:
I have just learned that "businessDate has been added for CoB job/tasks, it should not be used for floating rate setup" (Victor Romero). I will add this information to the Jira ticket and work on this another time.

PR #2392, opened 2025-04-08
feat: remove feature xbrl/mix reporting

steinwinde:
Removed the non-working XBRL and MIX report feature from the app

FIXES: WEB-45


IOhacker:
LGTM

PR #2391, opened 2025-04-08
docs: Update CONTRIBUTING, README

steinwinde:
Moved CONTRIBUTING file to root folder, updated CONTRIBUTING and README, e.g. adding release history section; referring to Jira and Slack instead of Github issues and Gitter; adding a section on Jira life cycle; and more

FIXES: WEB-123, WEB-124, WEB-125

PR #2393, opened 2025-04-08
build: Update version and bug tracker in package.json

steinwinde:
package.json updates: Version becomes 1.0.0, bug tracker becomes Jira

FIXES: WEB-127


steinwinde:
Ideally this would go live on Sunday, 13 April 2025 - see also #2391 .

IOhacker:
LGTM

PR #2395, opened 2025-04-09
fix: Fixes tab titles, reinstates top 100 activities code

steinwinde:
Fixes tab titles both in English and other languages.
From what I can tell #2376 worked only for 3 
pages - those 3 with translations in labels.titles.*. On 
the other hand, it did not even work for English on most pages.
In addition to that, #2376 removed code regarding 
top 100 user activities, without any explanation.
The PR here is meant to fix all of that and hopefully doesn't affect #2385.

FIXES: WEB-130, WEB-40


IOhacker:
LGTM

PR #2429, opened 2025-05-01
fix: broken HTML due to unterminated quotes

steinwinde:
This is (hopefully) the remaining bit necessary to correct all HTML issues covered by the ticket.

FIXES: WEB-150


IOhacker:
LGTM

PR #2430, opened 2025-05-01
fix: broken ngIf continuation

steinwinde:
3 broken *ngIf expressions that got outside of their elements in 2 files.

FIXES: WEB-150


IOhacker:
LGTM

PR #2431, opened 2025-05-01
fix: broken flex calc style directive

steinwinde:
There were quite a few broken element attributes like
   ="calc(90%-20px)"
which got replaced by CSS classes now.
FIXES: WEB-150


IOhacker:
LGTM

PR #2432, opened 2025-05-01
fix: Make it possible to navigate to the Centers page

steinwinde:
Adding some missing providers in CentersModule.
FIXES: WEB-158


IOhacker:
LGTM

PR #2438, opened 2025-05-03
fix: buttons with both mat-button and mat-icon-button spec

steinwinde:
Specifying both mat-button and mat-icon-button in a <button…> tag appears in markup of 2 files.

FIXES: WEB-163


IOhacker:
LGMT

PR #2441, opened 2025-05-04
fix: add missing mat-tab-nav-panels

steinwinde:
This adds mat-tab-nav-panels below nav mat-tab-nav-bars.

FIXES: WEB-166


IOhacker:
LGTM

PR #2443, opened 2025-05-04
fix: Implements themes for Angular 15

steinwinde:
This fixes the start-up warnings regarding the legacy themes we got after going for Angular 15. It implements the 
default theme and the selection of other color schemes.

FIXES: WEB-168


IOhacker:
LGTM

PR #2474, opened 2025-05-26
replace datetime-picker with stub and remove suppressImplicitAnyIndexErrors

gkbishnoi07:
## Description

- Introduced `DatetimePickerStubComponent` to replace `ngx-mat-datetime-picker` for now (temporary)
- Removed `suppressImplicitAnyIndexErrors` no longer needed with updated typings
- Updated `zone.js` import in `polyfills.ts` from `zone.js/dist/zone` to `zone.js`


IOhacker:
LGTM

PR #2475, opened 2025-05-27
Web 183 update angular version 16 to version 17

gkbishnoi07:
## Description

This change resolves all TypeScript errors that emerged after the removal of the suppressImplicitAnyIndexErrors compiler option. This removal is necessary because Angular 17 no longer supports or recommends relying on this flag, in line with stricter type enforcement and modern TypeScript best practices.

By explicitly addressing the previously suppressed any index errors, this update improves type safety and future-proofs the codebase for continued compatibility with Angular 17 and beyond.

No additional dependencies were introduced as part of this change.


gkbishnoi07:
sorry by mistake push

PR #2487, opened 2025-06-02
fix: client list no longer loading due to preloadClients default

gkbishnoi07:
## Description

This fix resolves an issue where the client list was not loading by default on the Clients page. Previously, users had to manually search for clients to see any results because the preloadClients flag was not set correctly. This update ensures that the client list now loads automatically when navigating to the Clients page, improving user experience and restoring expected functionality.

before
Screenshot 2025-06-02 215105

after
image


steinwinde:
The fix is based on existing code in the dev branch. But the assignment doesn't make much sense (and neither similar assignments in the same file). We will address these issues at a later stage.

gkbishnoi07:
> The fix is based on existing code in the dev branch. But the assignment doesn't make much sense (and neither similar assignments in the same file). We will address these issues at a later stage.

agreed

PR #2488, opened 2025-06-02
fix: Enables entry of datetime in Angular 17

steinwinde:
Angular 17 doesn't come with a datetime picker. We therefore need a workaround for entering date times. This provides such a workaround for the web app versions up to Angular 19, where we can combine date time pickers.

https://github.com/user-attachments/assets/a42dabf9-93d8-42c1-b583-1ccea47bc077

FIXES: WEB-187

### DIFF
--- a/src/app/shared/form-dialog/formfield/formfield.component.html
+++ b/src/app/shared/form-dialog/formfield/formfield.component.html
@@ -50,14 +50,12 @@
     <mat-label>{{ formfield.label }}</mat-label>
     <input
       matInput
-      [ngxMatDatetimePicker]="datetimePicker"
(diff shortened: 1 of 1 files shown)


steinwinde:
We go for https://www.npmjs.com/package/@ngxmc/datetime-picker up to version Angular 19 instead. I'm therefore closing this PR.

steinwinde:
It turns out that the fix pointed out in my previous comment doesn't work for Angular 17. Therefore I'm re-opening...

adamsaghy (approved):
Acceptable hence there is no "official" support for date time picker for Angular 17

PR #2501, opened 2025-06-08
Build: update package-lock.json, increase version of tailwind

steinwinde:
package-lock.json had not been updated for Angular 18 and hereby is;
slightly increased version of tailwindcss to make it compatible

FIXES: WEB-196


steinwinde:
Two commits instead of one. Closing.

PR #2504, opened 2025-06-08
Resolution to Web-109: Update gsim-account.component.html

mjpugh811:
This is the resolution to the ticket Web-109. Clicking on the checkbox on the GSIM Account Overview page at:

http://localhost:4200/#/groups/1/savings-accounts/gsim-account/000000004

would previously result in an 404 error. It now successfully redirects to an approval dialog at:

http://localhost:4200/#/groups/1/savings-accounts/4/actions/Approve


steinwinde (approved):
I've looked at the change, and this should fix it - thanks!

PR #2505, opened 2025-06-09
fix: Layout on fixed deposit and recurring deposit product

steinwinde:
Fixes layout of fixed deposit and recurring deposit product preview steps, plus a layout issue in the fixed deposit general tab.

FIXES: WEB-197


steinwinde:
Alberto has fixed these issues himself and merged his PR.

PR #2507, opened 2025-06-09
build: replace Karma with Jest everywhere

steinwinde:
Replaces Karma by Jest; removes all auto-generated Karma tests, but leaves (already before) commented custom tests in place; adjusts env variable for smoother runs of ng test.

FIXES: WEB-198


steinwinde:
I'm resolving the conflict in a new PR.

PR #2543, opened 2025-06-29
fix: Fixes many smaller layout issues in the read-only product views

steinwinde:
Fixes many smaller layout issues in the read-only product views:
loan, saving and share product

FIXES: WEB-219


gkbishnoi07 (approved):
LGTM

PR #2551, opened 2025-07-07
fix: revert font size to 14px

steinwinde:
The Angular updates from 14 to 19 have caused
the font size to change. This was not intended.

FIXES: WEB-247


gkbishnoi07 (approved):
LGTM

PR #2580, opened 2025-08-07
WEB-267 Remove tenant identifier for OAuth token requests

sidhantgoel:
## Description

The problem is Fineract-Platform-TenantId header is rejected by CORS policy on the endpoints exposed by keycloak.
This is a fix to remove this header from the requests to keycloack oauth server.

WEB-267

## Checklist

### DIFF
--- a/src/app/core/authentication/authentication.interceptor.ts
+++ b/src/app/core/authentication/authentication.interceptor.ts
@@ -36,6 +36,14 @@ export class AuthenticationInterceptor implements HttpInterceptor {
     if (this.settingsService.tenantIdentifier) {
       httpOptions.headers['Fineract-Platform-TenantId'] = this.settingsService.tenantIdentifier;
     }
+    if (
+      environment.oauth.enabled &&
+      (request.url.includes(`${environment.oauth.serverUrl}/token`) ||
+        request.url.includes(`${environment.oauth.serverUrl}/logout`))
+    ) {
+      // If the request is for OAuth token, we do not set the tenant identifier.
+      delete httpOptions.headers[authorizationTenantHeader];
+    }
     request = request.clone({ setHeaders: httpOptions.headers });
     return next.handle(request);
   }
--- a/src/app/core/authentication/authentication.service.ts
+++ b/src/app/core/authentication/authentication.service.ts
@@ -105,15 +105,12 @@ export class AuthenticationService {
       httpParams = httpParams.set('grant_type', 'password');
       let headers = new HttpHeaders();
       headers = headers.set('Content-Type', 'application/x-www-form-urlencoded');
-      return this.http
-        .disableApiPrefix()
-        .post(`${environment.oauth.serverUrl}/token`, httpParams.toString(), { headers: headers })
-        .pipe(
-          map((tokenResponse: OAuth2Token) => {
-            this.getUserDetails(tokenResponse);
-            return of(true);
-          })
-        );
+      return this.http.post(`${environment.oauth.serverUrl}/token`, httpParams.toString(), { headers: headers }).pipe(
+        map((tokenResponse: OAuth2Token) => {
+          this.getUserDetails(tokenResponse);
+          return of(true);
+        })
+      );
     } else {
       return this.http
         .post('/authentication', { username: loginContext.username, password: loginContext.password })
@@ -137,7 +134,6 @@ export class AuthenticationService {
     let headers = new HttpHeaders();
     headers = headers.set('Authorization', 'bearer ' + tokenResponse.access_token);
     this.http
-      .disableApiPrefix()
       .get(`${environment.serverUrl}/userdetails`, { headers: headers })
       .subscribe((credentials: Credentials) => {
         this.onLoginSuccess(credentials);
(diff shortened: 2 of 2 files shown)


steinwinde (changes requested):
Please create a Jira ticket, assign it to you, update its status and reference it in your PR description: https://mifosforge.jira.com/jira/software/c/projects/WEB/list?direction=DESC&sortBy=key This is a requirement for PRs. Thank you for contributing!

sidhantgoel:
@steinwinde Thanks for the update, please let me know if it looks ok now?

IOhacker:
@sidhantgoel I think this change oveheads the validation of any OIDC tool. Keycloak is a reference but there are others. Which is the benefit of removing the fineract tenant id header?

sidhantgoel:
@IOhacker In my opinion, we shouldn't send headers to the OIDC or any other server if they're not expected or used. Most likely, the provider will reject them. While CORS doesn't inherently block specific headers, the provider defines which headers are allowed. If a header isn't included in that list, the browser will block the request. Is there any downside to simply removing it?

IOhacker:
Cors is for browsers only, not for api calls.

You can see that the current code is working for oauth
https://oauth.mifos.community mifos and password are the credentials

The fineract tenant id is requiered for the Apache Fineract

Regards

Victor

sidhantgoel:
@IOhacker  CORS applies to API calls made through the browser—it's necessary for Fineract, but not for OAuth. So, there's no point in sending it with OAuth requests. I've only removed it for OAuth, not for Fineract. Also, just because it works with one auth provider doesn't guarantee it will work with others. We should adhere to the OIDC specification, and this header isn't defined there.

Just to clarify, this change won't remove the header from any Fineract requests—even when OAuth is enabled. It only removes the header from requests sent to the OAuth server by the web app.

IOhacker:
I think this change is not relevant. Because the header is not part of the
request. I have show you that it is working. There are changes like
managing the users and roles using the Mifos admin UI and these changes are
more relevant.

IOhacker:
Also the console.log must be removed.

sidhantgoel:
Sure will remove console.log but this header is part of token request and hence it is not working with keyclock. Please see the screenshot captured from the url shared.
Screenshot 2025-08-11 at 12 06 42 PM

IOhacker:
I dont see any error in the screen shared. You got a 200 which means that the flow is working. Even the screen shows the next api calls. This ticket is not showing any previous error.

IOhacker:
Here is the video of the instance working with all the endpoints (login, logout, refresh token) https://youtu.be/xXCCPsxqvR4 son then I don't think this change must be merged.

sidhantgoel:
t's not working with Keycloak. I'm not saying the instance itself is broken—I shared a screenshot to show that the header is indeed being sent to the OIDC server.

sidhantgoel:
And if you're saying it should only work with your configuration and that Keycloak or other providers don't need to be supported—well, so be it 😄

IOhacker:
This is using Keycloak. Please take some time to read the video description and my previous comments. Until now this change is not approved by me.

sidhantgoel:
Can you tell me which version ov Keycloak it is because it is definetly not working with latest version?

sidhantgoel:
Please read the issue already raised with Keycloak https://github.com/keycloak/keycloak/issues/12682

IOhacker:
It is using the latest Keycloak version 26.3.2-0

The issue is at another point. Please check the image. 

image

I insist this code change is not the solution.

sidhantgoel:
Can you point me towards the right fix? The header is being sent from the interceptor wich is common for oauth and fineract requests, should we not use interceptor in case of oauth requests?

IOhacker:
Take you time to debug it. But this code is not the solution.

sidhantgoel:
I fixed the issue that you have shown in the screenshot, the CORS issue came after that.

IOhacker:
The fix is not present in the code in the PR or maybe you missed
making the commit. This is just a header removal. By the way you must
squash and commit to all the PR sent to the repository.

sidhantgoel:
The issue here is that we're trying to augment `HttpClient`, which is provided via `provideHttpClient`. Because of this, the provider configuration we set up for creating `HttpService` doesn't take effect—`HttpClient` is always injected instead of `HttpService`, and `HttpClient` doesn't have the `disableApiPrefix` method.

Augmenting in this case either won’t work at all or would require a lot of hacks and/or copying code from the original module that provides `HttpClient`, which I don't recommend.

I didn’t commit that code because I simply commented out the `disableApiPrefix` call. The handling for full URLs is already covered by the `ApiPrefixInterceptor`, and we can always pass a full URL when we don’t want the API prefix added.

If required I can commit that code also.

IOhacker:
Do you have a video of your fix working?

sidhantgoel:
https://github.com/user-attachments/assets/ed3b51b9-21e0-4506-966e-5e93e4afdf64

Here is the screen recording

IOhacker:
It is possible to show the developer mode? I mean to view the keycloak
enpoints being called.

sidhantgoel:
https://github.com/user-attachments/assets/1111bd2e-8e85-408b-adc4-eae69a9b631c

Here you go

IOhacker:
Squash and commit the pr

sidhantgoel:
done, please check

src/app/core/authentication/authentication.interceptor.ts
```diff
@@ -36,6 +36,14 @@ export class AuthenticationInterceptor implements HttpInterceptor {
     if (this.settingsService.tenantIdentifier) {
       httpOptions.headers['Fineract-Platform-TenantId'] = this.settingsService.tenantIdentifier;
     }
+    if (
+      environment.oauth.enabled &&
+      (request.url.includes(`${environment.oauth.serverUrl}/token`) ||
+        request.url.includes(`${environment.oauth.serverUrl}/logout`))
+    ) {
```
IOhacker, line 43: .

IOhacker: The changes on this file are not required. Please remove from the PR

PR #2592, opened 2025-08-17
connect web-app to @fineract/client v1.0.4

gkbishnoi07:
## Description

- Install and pin @fineract/client to v1.0.4 (published from fineract-client-ts)
- Remove any local path/workspace overrides that previously pointed to packages/fineract-clients


IOhacker (approved):
LGTM

PR #2598, opened 2025-08-19
WEB-279 fix null reference error

jacobo-vargas-garcia:
## Description  

This PR fixes a runtime error that occurs in the Loan Product creation workflow when switching the loan schedule type from **Cumulative** to **Progressive**.  

The issue was caused by a null reference in the `LoanProductDeferredIncomeRecognitionStepComponent`, where the property `capitalizedIncome` was being accessed without verifying whether `deferredIncomeRecognition` was initialized.  

**Fix applied:**  
A null-safe access check (`?.`) was introduced to safely handle uninitialized objects and avoid runtime exceptions.  

This ensures that the UI behaves correctly and no errors are thrown when switching between schedule types.  

@alberto-art3ch @therajanmaurya @vidakovic @bharathcgowda 
Please can you help me with the revision, thank you.
## Screenshots 

**Before (error in console):**  

imagen

imagen


bharathcgowda:
@adamsaghy could you please help in reviewing this PR?

steinwinde (approved):
I haven't tested it, but the change looks 100% safe.

PR #2628, opened 2025-09-03
fix: Correct dependencies of Entity selection in Entity Data Table ch…

steinwinde:
…ecks

Status and other dependent dropdowns didn't show values dependent correctly on Entity selection.

This also fixes the misspelled name part "enity" to "entity".

FIXES: WEB-302


IOhacker (approved):
LGTM

PR #2676, opened 2025-09-22
WEB-76 Smaller text issues when editing Fixed Deposit Products

JaySoni1:
**Changes Made :-**

-Fixed typo in translation keys ("greterthan" to "greaterthan") across all language files.
-Added missing translation keys :- 'labels.inputs.Slab' and 'labels.buttons.this' to all translation files.
-Ensured consistent translations for "Slab" and "This" in all supported languages.

[WEB :- 76](https://mifosforge.jira.com/jira/software/c/projects/WEB/boards/62?assignee=712020%3A5685dc80-2da8-4d54-80e6-7b69c296926e&selectedIssue=WEB-76)


JaySoni1:
@steinwinde Please review .

steinwinde (changes requested):
Hi @jay, I'm sorry, but there must be something wrong:
- Regarding "greterthan", the only change made with this PR is an adjustment of the key in the language files. If the key is changed, its use must be adjusted accordingly. Where have you changed its use?
- You also add "labels.inputs.Slab". I think this would be good to have, but you added it inside of "labels": {"inputs": ... and then you write again "labels.inputs.Slab". Does the key not become labels.inputs.labels.inputs.Slab now?
- You also added "labels.buttons.this", but I don't see where this is used at all. 
All my comments would be nullified, if you had tested each change separately. Have you done so? We always have to test each change and if this test was successful in your case, please let me know.

JaySoni1:
@steinwinde  Thank you for your review . 

1. All references to the "greterthan" key in the codebase were updated to "greaterthan" to match the corrected translation key. This includes any component, service, or template where the key was used for displaying validation messages or labels.
2. The "labels.inputs.Slab" key was added directly within the "labels.inputs" section, so its full path is "labels.inputs.Slab" and not "labels.inputs.labels.inputs.Slab".
3. The "labels.buttons.this" key was added for consistency and for future use , but it is not currently used in the codebase.
4. I have tested  each change separately, and the UI displays the correct labels and translations.

steinwinde:
@JaySoni1 , I had a closer look at a few of your points:
1. The variable "greterthan" originates in Fineract source code. We can't correct its spelling, because otherwise the key in our language files can't be found. Please have a look here: https://github.com/apache/fineract/blob/develop/fineract-core/src/main/java/org/apache/fineract/portfolio/common/service/CommonEnumerations.java
2. The problem with "labels.inputs.Slab" is not fixed by your addition. I tested your addition and it adds indeed a translation key that would need to be referenced "labels.inputs.labels.inputs.Slab". The fact "labels.inputs.Slab" works as a reference is due to an existing translation key "Slab" below "labels", "inputs". Just have a look two lines above your addition. This is not a global key.
3. Please don't add translation keys to the translation files that are only for pure reasons of "consistency" or "future use". If you think this is important, please discuss in the web app channel and create a separate Jira ticket for this purpose.

PR #2711, opened 2025-10-18
Angular update 20

steinwinde:
This is a first attempt to bring the dev branch from Angular 19 to 20. All official chores to accomplish this have been done and "ng serve" starts without error (but with lots of warnings).

FIXES: WEB-350


alberto-art3ch (changes requested):
Too many Commits in the PR and huge amount of changes in the HTML, high risk to break the UI

steinwinde:
Rather than fixing this PR, I'm providing a new one

PR #2712, opened 2025-10-19
Replace a dependecy in version.js

steinwinde:
version.js had a dependency on fs-extra, which is
unnecessary, because we can use the built-in
Node functionality.

FIXES: WEB-351

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependency management by leveraging Node.js built-in modules instead of external packages, improving resource efficiency without user-facing changes.


IOhacker (approved):
LGTM

PR #2724, opened 2025-10-28
WEB-378-improve the css office-navigation-page

shubhamkumar9199:
## Description

Improved the CSS styling of the Office Navigation page to ensure proper alignment of icon.
This change enhances the visual consistency and layout across the navigation section, improving user experience.

#{Issue Number}
WEB-378
## Screenshots, if any
before:
Screenshot 2025-10-28 012441
after:
Screenshot 2025-11-01 231630

## Checklist

- [X] If you have multiple commits please combine them into one commit by squashing them.

- [ X ] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

## Summary by CodeRabbit

* **Style**
  * Enhanced responsive layouts: new flex/grid behavior, breakpoints at 768px and 1200px, improved spacing and small-screen adjustments.
  * Refined header and card typography with a stronger body style and adjusted sizing; improved dark-theme color overrides.

* **Refactor**
  * Moved external identifier out of the header into the card content for clearer presentation and simplified header structure.

### DIFF
--- a/src/app/navigation/navigation.component.scss
+++ b/src/app/navigation/navigation.component.scss
@@ -1,0 +1,54 @@
+:host {
+  display: block;
+}
+
+.container {
+  width: 100%;
+}
+
+.layout-row-wrap.responsive-column {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+}
+
+.flex-48 {
+  flex: 1 1 48%;
+  min-width: 320px;
+}
+
+mat-card {
+  padding: 16px;
+  border-radius: 8px;
+  overflow: hidden;
+}
+
+mat-card-content {
+  display: grid;
+  grid-template-columns: 100%;
+}
+
+mat-form-field {
+  width: 100%;
+}
+
+mat-label {
+  letter-spacing: 0.2px;
+}
+
+@media (768px <= width) {
+  mat-card-content {
+    grid-template-columns: 50% 50%;
+  }
+}
+
+@media (1200px <= width) {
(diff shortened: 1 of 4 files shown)


steinwinde (changes requested):
Sorry, but I don't like this fix for many reasons:
Labels are rendered in capitals only. Why? I don't see this in any similar forms in our app.
Underlines are rendered both below the labels and (separately) below their values. I neither see this in similar forms. 
We really should use the same CSS again and again, if possible - to the maximum extent possible.
Our forms are and should remain simple - in their CSS too. Why do we need CSS variables here? This looks like unnecessary overhead. Just declare two classes.
Whatever you do, please look how this is done elsewhere. E.g. if you introduce "transform: scale(x)", please make sure that this behaves exactly the same way it behaves elsewhere, e.g. not 1.05, but 1.1. (But I'd prefer we rather not do this at all.) E.g. if you introduce a colour "#212121", ask yourself, if we used the colour elsewhere - and if not, why should we use it here?! The user shouldn't see different colour schemes in each form, but one single colour scheme across all forms. I haven't checked your other CSS values, but they should not be specific for this given form.
The Jira-Ticket pointed out a layout bug. Please don't introduce fancy things that we can't do consistently across the application, because we already have a todo list that is too long.

shubhamkumar9199:
@steinwinde Sorry about that — this issue won’t happen again. If it appears in any of my previous PRs, I’ll review and correct it to make sure it doesn’t repeat

PR #2772, opened 2025-11-05
WEB-(410)-fix: correct type comparison in view-history-scheduler-job filter logic

shubhamkumar9199:
## Description
This PR fixes a type comparison bug in the view-history-scheduler-job component’s filter logic.
Previously, the code compared a string (val) with a number (parseInt(filter.value, 10)) using strict equality (===), which caused filtering by numeric values (like version numbers) to always return false

#{Issue Number}
WEB-(410)
## Checklist

- [X] If you have multiple commits please combine them into one commit by squashing them.

- [X] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filtering behavior: numeric comparisons are now applied only when both the filter and the item value are present and numeric; otherwise the filter falls back to a case-insensitive text match. This prevents incorrect numeric comparisons and yields more intuitive results when fields or filter inputs are empty or non-numeric.

### DIFF
--- a/src/app/system/manage-jobs/scheduler-jobs/view-history-scheduler-job/view-history-scheduler-job.component.ts
+++ b/src/app/system/manage-jobs/scheduler-jobs/view-history-scheduler-job/view-history-scheduler-job.component.ts
@@ -113,9 +113,9 @@ export class ViewHistorySchedulerJobComponent implements OnInit {
       const filters = JSON.parse(filtersJson);
       filters.forEach((filter: any) => {
         const val = data[filter.id] === null ? '' : data[filter.id];
-        if (filter.value !== '') {
-          matchFilter.push(val === parseInt(filter.value, 10));
-        } else if (filter.value === '') {
+        if (filter.value !== '' && val !== '') {
+          matchFilter.push(parseInt(val.toString(), 10) === parseInt(filter.value, 10));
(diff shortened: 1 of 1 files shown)


src/app/system/manage-jobs/scheduler-jobs/view-history-scheduler-job/view-history-scheduler-job.component.ts
```diff
@@ -114,7 +114,7 @@ export class ViewHistorySchedulerJobComponent implements OnInit {
       filters.forEach((filter: any) => {
         const val = data[filter.id] === null ? '' : data[filter.id];
         if (filter.value !== '') {
-          matchFilter.push(val === parseInt(filter.value, 10));
+          matchFilter.push(parseInt(val.toString(), 10) === parseInt(filter.value, 10));
```
alberto-art3ch, line 117: If the case the val variable has an empty value '' 
It is better to avoid that part, I mean It is better to validate this as part of the if condition

alberto-art3ch (changes requested):
Please see my comments

shubhamkumar9199:
@alberto-art3ch  i made the changes please review

alberto-art3ch (approved):
LGTM

PR #2747, opened 2025-11-03
WEB-(394)-fix: replace all direct JSON.parse usages with safe parsing utilities

shubhamkumar9199:
## Description

Replaced all direct JSON.parse() usages across the codebase with safe parsing utilities (safeParse, safeParseArray, safeParseObject) to prevent crashes from invalid or null JSON data.
#{Issue Number}
WEB-(394)

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [X ] If you have multiple commits please combine them into one commit by squashing them.

- [ X] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

## Release Notes

* **Bug Fixes**
  * Improved data handling and error resilience throughout the application. Settings, themes, date formats, user activities, and configuration data now safely process stored information with intelligent fallback mechanisms. The system gracefully handles malformed, incomplete, or missing data instead of crashing, preventing data loss and ensuring more stable, reliable, and consistent behavior.

PR #2798, opened 2025-11-17
Revert "WEB-(394)-fix: replace all direct JSON.parse usages with safe parsing utilities"

IOhacker:
Reverts openMF/web-app#2747 because incognito mode issues

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal data parsing logic by removing custom parsing utilities and adopting standard parsing methods across the application. This change affects how user preferences, themes, and configuration data are retrieved from storage. Settings retrieval behavior may change if stored data is malformed or missing.


IOhacker (commented):
PR for revert the incognito mode issue

steinwinde (approved):
This is a plain revert (I have checked) to the version that used to work.

PR #3058, opened 2026-01-28
WEB-216: UI allows to delete a group that is not "pending", but Fineract sends back error

parth-sharma-10:
[WEB-216](https://mifosforge.jira.com/jira/software/c/projects/WEB/boards/62?issueType=1&selectedIssue=WEB-216)
The delete group action was previously available in the UI for all group states, even though deletion is only allowed when a group is in the Pending state. This resulted in forbidden API calls and a confusing user experience.

This change disables the Delete action in the group actions menu when the group is not in a Pending state and adds a defensive guard in the component logic to prevent invalid delete attempts. This aligns the UI behavior with backend business rules and improves overall usability.

No additional dependencies are required for this change.

image

[WEB-216]: https://mifosforge.jira.com/browse/WEB-216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Release Notes

* **Bug Fixes**
  * Group deletion is now restricted to pending groups only; users attempting to delete other groups will see a disabled button with a clarifying tooltip.
  * Improved locale normalization in loan application reports for better accuracy across different regional settings.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/app/groups/groups-view/groups-view.component.html
+++ b/src/app/groups/groups-view/groups-view.component.html
@@ -198,7 +198,13 @@
           <button mat-menu-item *mifosxHasPermission="'CLOSE_GROUP'" (click)="doAction('Close')">
             {{ 'labels.buttons.Close' | translate }}
           </button>
-          <button mat-menu-item *mifosxHasPermission="'DELETE_GROUP'" (click)="doAction('Delete')">
+          <button
+            mat-menu-item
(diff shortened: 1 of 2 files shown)


alberto-art3ch (approved):
LGTM

PR #3130, opened 2026-02-11
WEB-694: Enforce one commit per pull request

heymanasvi:
Adds a GitHub Action to enforce a single commit per pull request.
The workflow fails pull requests that contain more than one commit, ensuring contributors squash commits before requesting review.

### DIFF
--- a/.github/workflows/one-commit-per-pr.yml
+++ b/.github/workflows/one-commit-per-pr.yml
@@ -1,0 +1,23 @@
+name: Enforce one commit per pull request
+
+on:
+  pull_request:
+
+jobs:
+  one-commit-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - run: |
+          git fetch origin ${{ github.base_ref }}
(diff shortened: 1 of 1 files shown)


IOhacker:
@heymanasvi the file is empty

heymanasvi:
@IOhacker Thanks for pointing that out. The file content has been updated now,it should be visible in the latest commit. Please let me know if anything looks incorrect.

DavidH-1 (changes requested):
This is not for community PR. the method of implementation will be by Mifos when the community has agreed if this is something it wants.

heymanasvi:
Thank you for the clarification @DavidH-1 .
Could you please guide me on the correct next step?
Should this be discussed on the mailing list or refined further before implementation?

DavidH-1:
yeah discussed on the web-app channel on slack.  I can already see differing opinions and no other project in mifos has this 1 commit rule so far so I'm not sure if they will really want it.

heymanasvi:
Got it, thanks for clarifying. Since there are differing views on this, I’ll pause work on it for now. I’m happy to revisit it later if the community decides to move forward. Rn I’ll pick up another issue.

PR #3204, opened 2026-02-22
WEB-746: Fix GSIM deposit and withdrawal routing duplication causing 404

heymanasvi:
While testing GSIM workflow, clicking on Deposit or Withdrawal
redirected to a malformed nested route:

#/groups/{groupId}/savings-accounts/gsim-account/{accountId}/actions/{action}

This resulted in a 404 error.

## Root Cause
The routerLink used relative navigation, causing the route
to remain nested under `gsim-account`.

## Fix
Updated routerLink to use correct navigation and extracted
groupId from route params to redirect properly to:

#/groups/{groupId}/savings-accounts/{accountId}/actions/{action}

## Testing

- Verified Deposit opens correctly
- Verified Withdrawal opens correctly
- No 404 occurs
- No undefined groupId errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Deposit and Withdrawal actions now navigate within the correct group context, preventing wrong/ambiguous account routes and improving navigation consistency.
  * Component now derives and uses the current group context for those actions so account operations open under the appropriate group.


heymanasvi:
Hii @IOhacker
While working on WEB-746 (Unable to generate collection sheet), I found a separate GSIM routing issue causing malformed nested routes and 404 errors during deposit/withdrawal.
This PR fixes only the GSIM routing issue. I’ll continue investigating the collection sheet problem and submit a separate PR once the root cause is identified.

IOhacker (approved):
Lgtm

PR #3330, opened 2026-03-06
WEB-819 Fix validation and translation issues in working capital product creation

JaySoni1:
**Changes Made :-** 

Changes Made:

-Updated validators from [Validators.min(1)] to [Validators.min(0.01)] for all period payment rate fields
 . Updated HTML min attributes and error messages to reflect 0.01 minimum .
 -Fixed Frequency Type error message to use correct translation key (labels.inputs.Frequency Type instead of [labels.inputs.Repaid every type] .
-Fixed NPV day count error message to use correct translation key (labels.inputs.Net Present Value day count instead of labels.inputs.NPV day count) .
-Made decimal places field empty by default for new Working Capital products .
-Added conditional logic to pass empty array for credit allocations in Working Capital products .
-Added null check before iterating credit allocation transaction types to prevent errors .

Before :-

image

image

image

After :-

image

image

image

### DIFF
--- a/src/app/products/loan-products/create-loan-product/create-loan-product.component.html
+++ b/src/app/products/loan-products/create-loan-product/create-loan-product.component.html
@@ -92,7 +92,7 @@
         >
         <mifosx-loan-product-payment-strategy-step
           [advancedPaymentAllocations]="advancedPaymentAllocations"
-          [advancedCreditAllocations]="advancedCreditAllocations"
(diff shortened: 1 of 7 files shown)


alberto-art3ch (approved):
LGTM

JaySoni1:
@alberto-art3ch Thank You for the review

PR #3340, opened 2026-03-09
WEB-826: Improve null-safety in ErrorHandlerInterceptor

tkshsbcue:
Description:
During network failures, CORS issues, or non-JSON server responses, response.error can be null or undefined. The interceptor directly accesses properties like developerMessage and errors, which can cause runtime errors.

Fix:
Use optional chaining and safe fallbacks when accessing response.error fields so the interceptor handles all error shapes gracefully.

ticket -> https://mifosforge.jira.com/browse/WEB-826?atlOrigin=eyJpIjoiMmRlZWE0NzNjNDkyNDQ0ZmJlMmRjMWYzMGE3NmJlZDkiLCJwIjoiaiJ9

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error extraction to handle missing or incomplete error data more safely.
  * Enhanced fallback logic so alerts display a meaningful message when expected error details are absent.
  * Adjusted logged message selection in non-production to prefer available response message with preserved status-based handling and alert behavior.

### DIFF
--- a/src/app/core/http/error-handler.interceptor.ts
+++ b/src/app/core/http/error-handler.interceptor.ts
@@ -45,11 +45,10 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
    */
   private handleError(response: HttpErrorResponse, request: HttpRequest<any>): Observable<HttpEvent<any>> {
     const status = response.status;
-    let errorMessage = response.error.developerMessage || response.message;
-    if (response.error.errors) {
-      if (response.error.errors[0]) {
-        errorMessage = response.error.errors[0].defaultUserMessage || response.error.errors[0].developerMessage;
-      }
+    const error = response?.error;
+    let errorMessage = error?.defaultUserMessage || error?.developerMessage || response?.message || 'Unknown error';
+    if (error?.errors?.[0]) {
+      errorMessage = error.errors[0].defaultUserMessage || error.errors[0].developerMessage || errorMessage;
     }
 
     const isClientImage404 = status === 404 && request.url.includes('/clients/') && request.url.includes('/images');
--- a/src/app/core/http/error-handler.interceptor.ts
+++ b/src/app/core/http/error-handler.interceptor.ts
@@ -46,7 +46,7 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
   private handleError(response: HttpErrorResponse, request: HttpRequest<any>): Observable<HttpEvent<any>> {
     const status = response.status;
     const error = response?.error;
-    let errorMessage = error?.defaultUserMessage || error?.developerMessage || response?.message || 'Unknown error';
+    let errorMessage = error?.defaultUserMessage || error?.developerMessage || 'Unknown error';
     if (error?.errors?.[0]) {
       errorMessage = error.errors[0].defaultUserMessage || error.errors[0].developerMessage || errorMessage;
     }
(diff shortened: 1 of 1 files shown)


IOhacker:
Run npx prettier

IOhacker (commented):
Run npx prettier

IOhacker (changes requested):
Run npx prettier

tkshsbcue:
@IOhacker fixed!

src/app/core/http/error-handler.interceptor.ts
```diff
@@ -45,11 +45,12 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
...
   private handleError(response: HttpErrorResponse, request: HttpRequest<any>): Observable<HttpEvent<any>> {
     const status = response.status;
-    let errorMessage = response.error.developerMessage || response.message;
-    if (response.error.errors) {
-      if (response.error.errors[0]) {
-        errorMessage = response.error.errors[0].defaultUserMessage || response.error.errors[0].developerMessage;
-      }
+    const error = response?.error;
+    let errorMessage =
+      error?.defaultUserMessage || error?.developerMessage || response?.message || 'Unknown error';
```
shubhamkumar9199: 'Unknown error' 
is hard-coded. If the application supports translations (which seems likely given other alert messages), it might be better to route this through the translation service or a shared constant to keep error messaging consistent.

src/app/core/http/error-handler.interceptor.ts
```diff
@@ -45,11 +45,12 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
    */
   private handleError(response: HttpErrorResponse, request: HttpRequest<any>): Observable<HttpEvent<any>> {
     const status = response.status;
-    let errorMessage = response.error.developerMessage || response.message;
-    if (response.error.errors) {
-      if (response.error.errors[0]) {
-        errorMessage = response.error.errors[0].defaultUserMessage || response.error.errors[0].developerMessage;
-      }
+    const error = response?.error;
+    let errorMessage =
```
shubhamkumar9199: In the previous implementation, only the defaultUserMessage from response.error.errors[0] was used.
In the new implementation, error?.defaultUserMessage is used first.

That changes the order of which message will be used to display to the user.
If this change was intentional, it might be worth mentioning this in the PR description so reviewers understand the reasoning.

src/app/core/http/error-handler.interceptor.ts
```diff
@@ -45,11 +45,12 @@ export class ErrorHandlerInterceptor implements HttpInterceptor {
    */
   private handleError(response: HttpErrorResponse, request: HttpRequest<any>): Observable<HttpEvent<any>> {
     const status = response.status;
-    let errorMessage = response.error.developerMessage || response.message;
-    if (response.error.errors) {
-      if (response.error.errors[0]) {
-        errorMessage = response.error.errors[0].defaultUserMessage || response.error.errors[0].developerMessage;
-      }
+    const error = response?.error;
```
shubhamkumar9199, line 48: response is typed as HttpErrorResponse and should never be null or undefined, so optional chaining here may not be necessary.
This could be simplified to const error = response.error;.

shubhamkumar9199:
@tkshsbcue Could you please squash the commits into a single commit

PR #3562, opened 2026-05-07
Removed other accounts

onesketchyguy:
Created a new web interface that only shows what we need here at WSCREDIT

## Release Notes

* **New Features**
  * Added EntityID field display for clients and groups across search results and account views.

* **Changes**
  * Restructured account navigation; loan accounts now display as primary view, with non-loan accounts separately toggled.
  * Removed savings account creation and management features.
  * Removed share account functionality and navigation tabs.
  * Updated search results to display EntityID instead of Account Number.


onesketchyguy:
Wrong fork, my bad

PR #3569, opened 2026-05-10
WEB-112: Sanitize GSIM fields before calculateLoanSchedule to avoid NPE

deepaksoni47:
## Description

This PR adds a frontend safeguard to prevent GSIM-linked fields from being sent to the backend loan schedule calculation endpoint, which was causing a 500 Internal Server Error.

**Changes Made:**
- Added `sanitizeLoanPayload()` private method to `LoansService` that recursively removes unsafe GSIM linkage fields (`linkAccountId`, `linkAccountOwnerId`, `linkAccountOwnerName`, `linkedSavingsAccount`) from loan payloads
- Applied sanitization in `calculateLoanSchedule()` endpoint before posting to `/loans?command=calculateLoanSchedule`
- Added comprehensive Jest unit tests to verify forbidden fields are removed at all nesting levels while allowed fields are preserved

**Why These Changes:**
The root cause of the 500 error is a `NullPointerException` in the Fineract backend (`LoanScheduleCalculationPlatformServiceImpl`) during GLIM + GSIM validation. While a permanent fix should be made in Apache Fineract, this frontend safeguard prevents invalid payloads from being sent, providing immediate relief for users.

**Dependencies:**
No new dependencies required. Uses existing Angular testing utilities (`provideHttpClient`, `provideHttpClientTesting` for Angular 20 compatibility).

[WEB-112](https://mifosforge.jira.com/browse/WEB-112)

## Screenshots, if any

N/A - Backend API safeguard (no UI changes)

## Checklist

- [x] Code follows project style (ran `npm run lint`, `npm run headers:add`)
- [x] Unit tests added and passing (`npm run test` - 2/2 tests passing)
- [x] Tested on dockerized Fineract backend - schedule generation no longer throws 500

[WEB-112]: https://mifosforge.jira.com/browse/WEB-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Release Notes

* **Bug Fixes**
  * Loan schedule calculation requests now filter out certain account linkage fields before sending to the server, ensuring cleaner request payloads.

* **Tests**
  * Added comprehensive unit tests for loan schedule calculation to verify proper request sanitization and field handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/web-app/pull/3569)

### DIFF
--- a/src/app/loans/loans.service.ts
+++ b/src/app/loans/loans.service.ts
@@ -30,6 +30,13 @@ export interface WorkingCapitalLoanDiscountUpdateRequest {
   providedIn: 'root'
 })
 export class LoansService {
+  private static readonly FORBIDDEN_LOAN_PAYLOAD_KEYS: ReadonlySet<string> = new Set([
+    'linkAccountId',
+    'linkAccountOwnerId',
+    'linkAccountOwnerName',
+    'linkedSavingsAccount'
+  ]);
+
   private http = inject(HttpClient);
   private settingsService = inject(SettingsService);
   private dateUtils = inject(Dates);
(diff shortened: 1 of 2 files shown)


deepaksoni47:
Addressed the CodeRabbit review comments and updated the implementation accordingly. Also re-ran tests and checks after the changes.

src/app/loans/loans.service.ts
```diff
@@ -30,6 +30,13 @@ export interface WorkingCapitalLoanDiscountUpdateRequest {
   providedIn: 'root'
 })
 export class LoansService {
+  private static readonly FORBIDDEN_LOAN_PAYLOAD_KEYS: ReadonlySet<string> = new Set([
```
IOhacker, line 33: what if a person requires to disburse a loan to a linked savings account?

IOhacker (changes requested):

src/app/loans/loans.service.ts
deepaksoni47: @IOhacker Thanks for pointing this out.

The intention here is only to sanitize payloads for the `calculateLoanSchedule()` flow, which is used for repayment schedule preview/calculation and does not perform actual loan disbursement operations.

Actual disbursement to linked savings accounts goes through separate actions/endpoints such as `disburse` and `disbursetosavings`, so those flows remain unaffected by this change.

The issue I observed was specific to the schedule calculation endpoint, where the backend was throwing an NPE when GSIM-linked fields were included in the calculation payload. Since these fields are not required for repayment schedule computation itself, this change removes them only for that calculation request as a temporary defensive workaround until the backend-side validation is addressed.

Please let me know if there's a specific disbursement scenario that uses `calculateLoanSchedule()` that I may have missed.

PR #3573, opened 2026-05-12
WEB-579: make API gateway tenant header name configurable

deepaksoni47:
## Description

Make the API Gateway (Mifos Endpoint) tenant header name configurable so different deployments (or API gateways) can use a custom header name without code changes.

- Add `apiGatewayHeaderName` to runtime environment defaults (default: `Fineract-Platform-TenantId`).
- Update `TokenInterceptor` to read `environment.apiGatewayHeaderName` and use it when setting the tenant header on outgoing requests.
- Preserve existing behavior by default to remain backward compatible with tests and deployments.

[WEB-579](https://mifosforge.jira.com/browse/WEB-579)

## Deployment / Ops notes

- To override the header name at runtime set `apiGatewayHeaderName` in your `window.env` or deployment environment (same mechanism used for other `loadedEnv` values).
- No DB or backend migration required.

## Screenshots, if any

N/A

## Checklist

[WEB-579]: https://mifosforge.jira.com/browse/WEB-579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Made the API gateway tenant header name configurable via environment with validation and a safe default, and ensured the app consistently uses this configured header.
  * Explicitly typed the environment configuration to include the new setting.
  * Minor cleanup: tenant selector trimming/filtering improved for more reliable display.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/web-app/pull/3573)

### DIFF
--- a/src/app/zitadel/token.interceptor.ts
+++ b/src/app/zitadel/token.interceptor.ts
@@ -19,11 +19,15 @@ export class TokenInterceptor implements HttpInterceptor {
 
   public environment = environment;
   FINERACT_PLATFORM_TENANT_IDENTIFIER = environment.fineractPlatformTenantId;
+  // Name used to send the Fineract tenant identifier header.
+  // The runtime key is `apiGatewayHeaderName` to allow gateway/runtime injection,
+  // but this value represents the Fineract tenant header name.
+  FINERACT_TENANT_HEADER_NAME = environment.apiGatewayHeaderName || 'Fineract-Platform-TenantId';
 
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
     const token = this.authService.getAccessToken();
     let headersConfig: { [key: string]: string } = {
-      'Fineract-Platform-TenantId': this.FINERACT_PLATFORM_TENANT_IDENTIFIER,
+      [this.FINERACT_TENANT_HEADER_NAME]: this.FINERACT_PLATFORM_TENANT_IDENTIFIER,
       'Content-Type': req.headers.get('Content-Type') || 'application/json'
     };
     const publicEndpoints = [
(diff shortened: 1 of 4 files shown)


IOhacker (changes requested):
Could you please elaborate a use case? The API Gateways can do any header change. In fact the sanbox is using an API Gateway.

src/app/zitadel/token.interceptor.ts
```diff
@@ -19,11 +19,12 @@ export class TokenInterceptor implements HttpInterceptor {
 
   public environment = environment;
   FINERACT_PLATFORM_TENANT_IDENTIFIER = environment.fineractPlatformTenantId;
+  API_GATEWAY_HEADER_NAME = environment.apiGatewayHeaderName || 'Fineract-Platform-TenantId';
```
IOhacker: This is not an API Gatewar Header, this is Apache Fineract Header

IOhacker (changes requested):

deepaksoni47:
> Could you please elaborate a use case? The API Gateways can do any header change. In fact the sanbox is using an API Gateway.

Thanks — good point.

The header we send is still the Fineract tenant header (`Fineract-Platform-TenantId` by default). We made its *name* configurable at runtime so deployments behind gateways/reverse proxies can use different tenant headers (`X-Tenant-Id`, `X-Client-Tenant`, etc.) without rebuilding the frontend or adding extra gateway translation rules.

This helps with multi-tenant routing, legacy gateway requirements, and QA/staging environments that already expect different header names.

I also renamed the in-code constant to `FINERACT_TENANT_HEADER_NAME` to make the intent clearer while keeping the runtime env key for backward compatibility.

IOhacker:
Those actions are done by a API Gateway. I think this ticket should be
closed. I will open a new one, because it is missing in Mifos WebApp the
header for the API Key or Subscription Key.

deepaksoni47:
Understood, thanks for the clarification. The intention was mainly to reduce deployment-specific gateway translation rules while keeping the default behavior unchanged.

Thanks also for pointing out the API Key / Subscription Key header support gap.

IOhacker:
As I have mentioned before, it is already done. And it is a header required
by Apache Fineract, check the sandbox.mifos.community deployment. ;)

PR #3598, opened 2026-05-21
WEB-956: share auth session across tabs and preserve deep links

jonattan-infante:
## Summary

Fixes [WEB-956](https://mifosforge.jira.com/browse/WEB-956).

When a logged-in user opens a Mifos URL in a new browser window (a link
from email/Slack, a bookmark, `target="_blank"`, etc.), the Web App
currently forces them to log in again — and then drops them on the home
dashboard instead of the URL they originally requested. The root cause
is twofold: credentials live in `sessionStorage` (per-tab), and the
authentication guard / login component discard the target URL.

This PR adopts the pattern used by
[`@supabase/auth-js`](https://github.com/supabase/auth-js/blob/main/src/GoTrueClient.ts)
and
[`@auth0/auth0-spa-js`](https://github.com/auth0/auth0-spa-js/blob/main/src/cache/cache-localstorage.ts):
`localStorage` as the single source of truth for the session, plus a
`BroadcastChannel` for cross-tab synchronisation. It also makes the
authentication guard preserve the target URL through the login flow.

### Changes

- **`authentication.service.ts`** — credentials are now persisted in
  `localStorage` unconditionally, so the authenticated session is visible
  to every tab/window of the same origin. The `rememberMe` flag is
  preserved for the backend token-expiration policy but no longer chooses
  the `Storage` instance.
- **`authentication.service.ts`** — a `BroadcastChannel` named
  `mifosXAuth` is created lazily (with a feature-detect fallback). Login
  broadcasts `{ type: 'login' }` and logout broadcasts `{ type: 'logout' }`.
  Other tabs rehydrate or wipe their state in real time, with no reload
  needed.
- **`authentication.guard.ts`** — `canActivate` now receives
  `RouterStateSnapshot` and forwards `state.url` as a `returnUrl` query
  parameter when redirecting to `/login`. Trivial targets (`/`, `/login`,
  `/login?...`) are skipped to avoid noise and redirect loops.
- **`login.component.ts`** — after a successful authentication, the
  component reads `returnUrl` from the activated route's `queryParamMap`
  and navigates there via `navigateByUrl`, falling back to `/` when
  absent.
- **`authentication.guard.spec.ts`** — adds 6 unit tests: authenticated
  pass-through, deep-link capture, loop avoidance for `/` and `/login`,
  stale-params handling, and URLs with query params + fragments.

### Before / after

| Scenario                                | master | this PR |
| --------------------------------------- | :----: | :-----: |
| click link while logged in (new window) |   ❌   |   ✅    |
| `returnUrl` preserved in `/login` URL   |   ❌   |   ✅    |
| deep link restored after login          |   ❌   |   ✅    |
| logout broadcasts across tabs           |  n/a   |   ✅    |

```
master   /#/clients  →  /#/login                          →  /#/home    (bug)
this PR  /#/clients  →  /#/login?returnUrl=%2Fclients     →  /#/clients
```

## Test plan

- [x] Unit tests pass:
      `npx jest src/app/core/authentication/authentication.guard.spec.ts`
- [x] Production build succeeds: `npm run build`
- [x] End-to-end validation in an isolated docker-compose stack
      (`postgres:17` + Fineract + `openmf/web-app:master` vs local build)
      using Playwright. All four scenarios in the table above pass.
- [ ] Reviewer manual smoke test:
  - [ ] Login → open `/#/clients` in a new window → no re-login.
  - [ ] Incognito → open `/#/clients` → land back on `/#/clients` after login.
  - [ ] Logout in tab A → tab B kicked out when navigating to a protected route.
  - [ ] OIDC / OAuth2 flows continue to work (no regressions in
        `restoreSession` for non-Basic auth modes).
- [ ] Reviewer verifies the `Remember Me` checkbox still controls the
      backend's `remember` flag (it does — only the client-side `Storage`
      choice was changed).

## Notes for reviewers

- This is a frontend-only change. No Fineract or schema changes.
- `sessionStorage` is still cleared on logout for safety (both
  `localStorage` and `sessionStorage` are wiped — see `setCredentials`),
  so users that had a stale `sessionStorage` from previous versions are
  not affected after upgrading.
- The `BroadcastChannel` feature-detect (`typeof BroadcastChannel !== 'undefined'`)
  keeps the service usable in environments lacking the API (very old
  browsers, some SSR contexts).
- I deliberately did NOT migrate to HTTP-only cookies (the approach used
  by NextAuth/Vercel) because it would require backend changes on
  Fineract and is out of scope for this fix.

## Summary by CodeRabbit

* **New Features**
  * Post-login restores the exact requested page (preserving query strings and fragments).
  * Credentials and 2FA tokens are persisted to localStorage and login/logout sync across open tabs for cross-tab session restoration.

* **Bug Fixes**
  * Redirect logic avoids returnUrl for home/login to prevent redirect loops and uses replace navigation on redirects.
  * Login flow now respects preserved returnUrl on successful authentication.

* **Tests**
  * Added comprehensive guard tests covering redirects, deep-link preservation (including login-history), and complex URL scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openMF/web-app/pull/3598?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

[WEB-956]: https://mifosforge.jira.com/browse/WEB-956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/auth.setup.ts
+++ b/playwright/auth.setup.ts
@@ -29,16 +29,23 @@ setup('authenticate', async ({ page, browser }) => {
   await loginPage.navigate();
   await loginPage.loginAndWaitForDashboard(username, password);
 
-  console.log('Auth setup: copying mifosXCredentials from sessionStorage → localStorage');
-  const credsCopied = await page.evaluate(() => {
-    const creds = sessionStorage.getItem('mifosXCredentials');
-    if (!creds) return false;
-    localStorage.setItem('mifosXCredentials', creds);
-    return true;
+  // Ensure mifosXCredentials is persisted in localStorage so storageState
+  // captures a shareable, cross-tab session. AuthenticationService now
+  // writes to localStorage directly, but we still accept sessionStorage
+  // as a legacy source while older branches/builds are around.
+  const credsAvailable = await page.evaluate(() => {
+    const fromLocal = localStorage.getItem('mifosXCredentials');
+    if (fromLocal) return true;
+    const fromSession = sessionStorage.getItem('mifosXCredentials');
+    if (fromSession) {
+      localStorage.setItem('mifosXCredentials', fromSession);
+      return true;
+    }
+    return false;
   });
 
-  if (!credsCopied) {
-    throw new Error('CRITICAL: mifosXCredentials not found in sessionStorage. ' + 'Did the auth storage key change?');
+  if (!credsAvailable) {
+    throw new Error('CRITICAL: mifosXCredentials not found in either storage after login.');
   }
 
   await page.context().storageState({ path: authFile });
--- a/src/app/core/authentication/authentication.guard.spec.ts
+++ b/src/app/core/authentication/authentication.guard.spec.ts
@@ -1,0 +1,95 @@
+/**
+ * Copyright since 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { ActivatedRouteSnapshot, Router, RouterStateSnapshot } from '@angular/router';
+
+import { AuthenticationGuard } from './authentication.guard';
+import { AuthenticationService } from './authentication.service';
+
+describe('AuthenticationGuard', () => {
+  let guard: AuthenticationGuard;
+  let authServiceMock: jest.Mocked<Pick<AuthenticationService, 'isAuthenticated' | 'logout'>>;
+  let routerMock: jest.Mocked<Pick<Router, 'navigate'>>;
+
+  const buildState = (url: string): RouterStateSnapshot => ({ url }) as RouterStateSnapshot;
+  const emptyRoute = {} as ActivatedRouteSnapshot;
+
+  beforeEach(() => {
+    authServiceMock = {
+      isAuthenticated: jest.fn(),
+      logout: jest.fn()
+    };
+    routerMock = { navigate: jest.fn() };
+
+    TestBed.configureTestingModule({
+      providers: [
+        AuthenticationGuard,
+        { provide: AuthenticationService, useValue: authServiceMock },
+        { provide: Router, useValue: routerMock }]
+    });
+
+    guard = TestBed.inject(AuthenticationGuard);
+  });
+
+  it('allows the route when user is authenticated', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(true);
+    const result = guard.canActivate(emptyRoute, buildState('/clients'));
+    expect(result).toBe(true);
+    expect(routerMock.navigate).not.toHaveBeenCalled();
+    expect(authServiceMock.logout).not.toHaveBeenCalled();
+  });
+
+  it('redirects to /login with returnUrl when target is a deep link', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(false);
+    const result = guard.canActivate(emptyRoute, buildState('/clients/1/general'));
+    expect(result).toBe(false);
+    expect(authServiceMock.logout).toHaveBeenCalledTimes(1);
+    expect(routerMock.navigate).toHaveBeenCalledWith(['/login'], {
+      queryParams: { returnUrl: '/clients/1/general' },
+      replaceUrl: true
+    });
+  });
+
+  it('redirects to /login WITHOUT returnUrl when target is "/" (default landing)', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(false);
+    guard.canActivate(emptyRoute, buildState('/'));
+    expect(routerMock.navigate).toHaveBeenCalledWith(['/login'], { queryParams: {}, replaceUrl: true });
+  });
+
+  it('redirects to /login WITHOUT returnUrl when target is already /login (avoids loops)', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(false);
+    guard.canActivate(emptyRoute, buildState('/login'));
+    expect(routerMock.navigate).toHaveBeenCalledWith(['/login'], { queryParams: {}, replaceUrl: true });
+  });
+
+  it('redirects to /login WITHOUT returnUrl when target is /login with stale params', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(false);
+    guard.canActivate(emptyRoute, buildState('/login?returnUrl=/foo'));
+    expect(routerMock.navigate).toHaveBeenCalledWith(['/login'], { queryParams: {}, replaceUrl: true });
+  });
+
+  it('preserves complex URLs with query params and fragments', () => {
+    authServiceMock.isAuthenticated.mockReturnValue(false);
+    const target = '/loans/42?tab=schedule#repayment';
+    guard.canActivate(emptyRoute, buildState(target));
+    expect(routerMock.navigate).toHaveBeenCalledWith(['/login'], {
+      queryParams: { returnUrl: target },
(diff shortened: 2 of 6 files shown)


jonattan-infante:
Thanks @coderabbitai — all three findings were valid. Addressed in the squashed commit (force-push):

- **`getSavedCredentials()` read order** — `localStorage` is now the authoritative source. A one-time migration promotes any leftover `sessionStorage` credentials so existing users keep their session across the upgrade.
- **Cross-tab logout 2FA cleanup** — the `BroadcastChannel` `logout` handler now also clears `mifosXTwoFactorAuthenticationToken` from both storages and calls `removeTwoFactorAuthorization()` on the interceptor, mirroring what `logout()` does on the active tab.
- **Login broadcast never fired** — replaced `wasLoggedIn = userLoggedIn$.getValue()` with `hadPersistedCredentials = !!getSavedCredentials()`. Because `onLoginSuccess()` flips `userLoggedIn$` to `true` *before* calling `setCredentials()`, the previous check always saw `true` and skipped the broadcast. The new check reads from storage and reliably announces fresh logins to other tabs.

I also squashed the two commits into a single commit per project convention.

jonattan-infante:
Thanks again @coderabbitai — second pass also valid. Addressed in the amended commit:

1. **Guard treats `/login-history` as a deep link** — replaced `targetUrl.startsWith('/login')` with an explicit check for the exact `/login` path (plus its `?…` / `#…` variants). Unrelated routes like `/login-history` now keep their `returnUrl`. Added a unit test for it (`/login-history`), test count is now 7.
2. **Legacy 2FA token migration** — `getSavedCredentials()` now also promotes `mifosXTwoFactorAuthenticationToken` from `sessionStorage` to `localStorage` together with the credentials, so Basic+2FA users coming from an older build don't end up with a stranded 2FA header.
3. **2FA token persisted before login broadcast** — in `onOTPValidateSuccess()`, the token is now written to `localStorage` *before* `setCredentials()` runs, so any other tab waking up from the cross-tab `{ type: 'login' }` broadcast rehydrates with a complete auth context (credentials + 2FA header).

jonattan-infante:
Quick follow-up @coderabbitai — I removed the `sessionStorage → localStorage` migration block from `getSavedCredentials()` after a second look. Reasoning:

- The only users it would help are those who were *actively logged in* at the exact moment of the deploy and refresh the page immediately. Anyone idle for more than the session timeout already has empty `sessionStorage`, and once the new bundle is served `setCredentials` always writes to `localStorage` going forward.
- The worst-case outcome without the migration is a single extra login for that narrow user window — about three seconds.
- The migration adds permanent cognitive load to a hot path in a security-sensitive file, in exchange for a one-time benefit that becomes dead code within minutes of the deploy.

I'm trading specific correctness here for long-term simplicity. Happy to revisit if a maintainer prefers the migration kept — it's a small block to put back. The 2FA migration suggestion (CodeRabbit #3284700933) is removed for the same reason; it only existed to support the credentials migration.

Force-pushed `c6a651b`. The other two fixes from the second pass (strict `/login` match in the guard and 2FA-token-before-broadcast in `onOTPValidateSuccess`) are unchanged.

jonattan-infante:
## Local CI parity run — `25 / 26` Playwright specs passing

I replicated the CI workflows against an isolated docker stack to validate this PR end-to-end. Results:

| Check | Status | Notes |
|---|:---:|---|
| Single commit (`commit-lint.yml`) | ✅ | 1 commit (`214d114`) |
| `npm run lint` (eslint + stylelint + prettier + htmlhint) | ✅ | clean |
| `headers:check` (`license-headers.yml`) | ✅ | all changed files have MPL-2.0 header |
| Unit tests (jest) | ✅ | 7/7 specs in `authentication.guard.spec.ts` |
| `npm run build:prod` | ✅ | bundle generated under budget |
| Playwright E2E (`playwright.yml`) | ✅ 25/26 | see below |

### About `playwright/auth.setup.ts`

This file currently reads credentials from `sessionStorage` and copies them to `localStorage` — a workaround that *exists because of the bug this PR fixes*. With `AuthenticationService` now writing to `localStorage` directly, the workaround would `throw 'mifosXCredentials not found in sessionStorage'` and break all authenticated specs.

I adapted it to a backwards-compatible check: prefer `localStorage` (post-fix), fall back to `sessionStorage` (pre-fix). No external behaviour change.

### About the 1 failing E2E spec

`close-client.spec.ts › should keep the client active when cancel or browser back is used from the close flow` — fails consistently with `element was detached from the DOM, retrying` on `actionsMenuItem.hover()`. The interaction is unrelated to authentication (it's a Material Design 3 menu hover/click on the client view). I suspect it's environment-specific seed data sensitivity rather than a regression from this PR. Happy to dig in further if a maintainer thinks otherwise.

Force-pushed `214d114`.

IOhacker (changes requested):
Which authentication methods were used for the testing? 2FA? OAUTH? Basic? OIDC?

jonattan-infante:
Both new findings from @coderabbitai are valid. Addressed in `9f2544c`:

### 1. Cross-tab OAuth/OIDC logout & storage consistency (Major)

- **Cleared OAuth library tokens in the cross-tab logout handler.** For non-Basic auth, `isAuthenticated()` relies on `oauthService.hasValidAccessToken()`. The handler now invokes `this.oauthService.logOut(true)` (no redirect) when `authMode !== Basic`, so a passive tab no longer reports a stale session after another tab logs out.
- **Removed the storage split in `initializeOAuthService()`.** It used to pick `sessionStorage` when `enableRememberMe` was false, while `setCredentials()` later forced `localStorage`. Half of the OAuth tokens ended up in the wrong store and `oauthService.logOut()` would only clean one of them. `initializeOAuthService()` now uses `localStorage` from the start, consistent with the rest of the auth flow.

### 2. Validate `returnUrl` before navigating after login (Minor)

`LoginComponent` now rejects unsafe `returnUrl` values before calling `navigateByUrl`. The check requires:

- value starts with `/` (relative path)
- value does not start with `//` (protocol-relative URLs → would escape origin)
- value is not exactly `/login` nor `/login?…` / `/login#…` (would bounce back to login)

Anything else falls back to `/`.

### Local pipeline parity (same as previous run)

| Check | Result |
|---|:---:|
| Single commit (`commit-lint.yml`) | ✅ |
| `npm run lint` | ✅ |
| `headers:check` | ✅ |
| Unit tests (7/7) | ✅ |
| `npm run build:prod` | ✅ |
| Playwright E2E | ✅ 25/26 (same unrelated `close-client` flake as before) |

Force-pushed `9f2544c`.

jonattan-infante:
Hi @IOhacker — I covered Basic end-to-end. The other three modes I couldn't fully validate E2E and I'll explain why below.

Setup: `apache/fineract:latest` upstream, `postgres:18.3-alpine`, and `ghcr.io/navikt/mock-oauth2-server:3.0.1` (same mock your CI uses), nginx serving this PR's build.

**Basic:** full Playwright suite passes against the upstream stack.

**OAuth2:** asserted that the cross-tab logout properly clears the OAuth library tokens and that the storage choice is consistent. Token issuance and discovery endpoints work end-to-end, but a real login flow hits an upstream Fineract bug (`SpelEvaluationException: getTenantIdentifier() on null`) — reproducible on stock `apache/fineract:latest` without this PR, happy to file separately.

**OIDC:** same code path as OAuth2 in the web-app, covered the same way. Same upstream bug gates the live flow.

**2FA:** verified the cross-tab logout also tears down the 2FA state, and that the 2FA token is persisted before the login broadcast fires so passive tabs rehydrate complete. No live E2E since the upstream image doesn't ship a 2FA tenant.

If you have a known-good OAuth/2FA environment to point at, I'll re-run there.

PR #3638, opened 2026-06-08
WEB-951 The 'Print' button should be visible 

JaySoni1:
**Chnages Made :-**

-Restore print button visibility condition when MIFOS_PRODUCTION_MODE is enabled .

[WEB-951](https://mifosforge.jira.com/browse/WEB-951)

[WEB-951]: https://mifosforge.jira.com/browse/WEB-951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the loan screen reports UI so the **Print** button is shown only in **production mode** (while keeping the same disabled/interaction behavior as before).

### DIFF
--- a/src/app/loans/loans-view/loan-account-actions/loan-screen-reports/loan-screen-reports.component.html
+++ b/src/app/loans/loans-view/loan-account-actions/loan-screen-reports/loan-screen-reports.component.html
@@ -45,10 +45,12 @@
 <div #output class="container">
   <mat-card class="layout-column gap-3percent">
     <div class="layout-align-end">
-      <button mat-stroked-button color="primary" [disabled]="!template" (click)="print()">
-        <fa-icon icon="file" class="m-r-10"></fa-icon>
-        {{ 'labels.buttons.Print' | translate }}
-      </button>
+      @if (productionMode) {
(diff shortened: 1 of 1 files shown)


alberto-art3ch (changes requested):
Which is the reason to have this constraint related to production env ?

IOhacker (dismissed):
LGTM. The constrain is to avoid any confusion with other users that don't use the Pentaho plugin.

alberto-art3ch:
> LGTM. The constrain is to avoid any confusion with other users that don't use the Pentaho plugin.

Two points:
If that is the rule, It has nothing to use productiveMode, and
Second, the current implementation is not correct, I mean *ngIf is not used more

JaySoni1:
@alberto-art3ch @IOhacker I have updated these PR please review

alberto-art3ch (approved):
LGTM

JaySoni1:
@alberto-art3ch @IOhacker Thank You for the review

PR #3724, opened 2026-07-16
fix(WEB-1038): include template id on update

AnvayKharb:
## Summary 

Includes the existing template ID in the update request payload when editing a template.

The current template update API requires the template ID in both the request path and request body. The Web App previously supplied it only in the path, causing template updates to fail with an `id.not-null` validation error.

The template creation flow remains unchanged.

#WEB-1038

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed template editing so the existing template ID is preserved when saving changes.

* **Tests**
  * Added unit test coverage for the template create/edit flows, verifying the correct save action is called and that submitted data matches expectations for both modes.


IOhacker (approved):
LGTM

PR #3736, opened 2026-07-20
fix(WEB-1049): prevent zero savings transaction amounts

AnvayKharb:
## Summry 

Prevents users from entering or submitting a zero transaction amount for regular savings Deposit and Withdrawal actions.

This change adds frontend validation to require a positive transaction amount before users can proceed to the confirmation step, aligning the Web-App behavior with the existing backend validation. The minimum valid amount is derived from the account currency precision, and focused unit tests have been added to cover the validation scenarios.

#WEB-1049

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced the minimum allowed transaction amount for deposits and withdrawals based on the account currency’s decimal precision.
  * Prevented users from proceeding to confirmation when the transaction amount is invalid (including zero, negative, and empty inputs).
  * Preserved validation that rejects amounts with excessive decimal precision.

* **Tests**
  * Added component test coverage for minimum amount behavior across currency precision and for confirmation-step prevention with invalid amounts.


IOhacker (approved):
LGTM

PR #3857, opened 2026-08-18
fix(notes): render note changes without a manual page refresh

parth-sharma-10:
## Description

**Problem**

Notes added, edited or deleted from a notes tab were not reflected in the list until the browser was refreshed manually. From the user's point of view there was no confirmation that the note had been saved, which invites capturing the same note twice. Reported against Client Notes (`/#/clients/94/notes` on the demo instance); the same defect is present on four other tabs.

**Root cause**

The notes tab components and the shared `EntityNotesTabComponent` that renders the list are both `ChangeDetectionStrategy.OnPush`, and the tabs updated `entityNotes` in place (`push()` / `splice()` / `entityNotes[index].note = …`).

1. The form submit is handled in the shared component's template, so the click marks the views dirty — the POST itself always worked.
2. The response arrives later. `ApplicationRef.tick()` skips OnPush views that are neither dirty nor have changed inputs. The parent is not dirty at that point, so it is skipped, `[entityNotes]` is never re-evaluated, and the child's `@for` never re-runs.
3. Reloading the page re-runs the route resolver, which is why a refresh appeared to fix it.

**Impacted features**

- Client Notes
- Loan Notes
- Savings Account Notes
- Center Notes
- Group Notes

Jira covers Client, Loan and Savings Account Notes. **Center and Group Notes were not separately reported but carry the identical defect**, so they are fixed here rather than left behind.

**Fix**

- `entityNotes` is now updated immutably — spread on add, `filter()` on delete, `map()` with a new object on edit — so the child's input reference actually changes.
- `ChangeDetectorRef.markForCheck()` is called after each successful response so the parent view is checked and the new reference reaches the binding.

Both halves are required. A new array reference alone never reaches the binding, because the OnPush parent is never checked after an HTTP response. `markForCheck()` alone leaves the child clean, because its input reference did not change.

- **Latent `bind(this)` defect.** Loans and savings passed no bound callbacks to the shared component, and groups bound only `addNote`. Since the shared component invokes them as `this.callbackAdd(...)`, those methods executed with `this` bound to `EntityNotesTabComponent` — which happens to inject the same entity services and hold `entityNotes` as an `@Input`, so the old mutating code limped along by accident while silently recording `createdByUsername: undefined` on new notes. With this change those callbacks use an injected `ChangeDetectorRef`, so leaving them unbound would throw. They are now bound the way clients and centers already did it.

No signals, no change to the shared component's public API, no change to service behaviour, no new dependencies.

**Regression coverage**

`notes-tab-change-detection.spec.ts` drives all five tabs through the DOM — form submit to add, action buttons with a stubbed `MatDialog` to edit and delete — and asserts the rendered `.note-card` list.

Two details make it a real regression test rather than one that passes either way:

- Change detection runs through `ApplicationRef.tick()` with the fixture's host view attached, which is the runtime path that skips clean OnPush views. `fixture.detectChanges()` force-refreshes the root view and hides the defect.
- The service mocks return a pending `Subject`, not `of(...)`. A synchronous mock resolves inside the event handler while the view is still dirty, and passes against the buggy code.

Reverting the component changes and keeping the spec turns all 15 tests red.

Jira: Client Notes, Loan Notes and Savings Account Notes tickets (IDs to be added). Center Notes and Group Notes have no ticket — the same defect was found in them while fixing the reported three.

## Screenshots, if any

**Before / After**

Client Notes behavior before and after the change: `client_notes_fixed.mp4`

This video is evidence for **Client Notes only**. Loan, Savings Account, Center and Group Notes are covered by the automated regression tests, not by this recording.

## Verification

| Check | Result |
|---|---|
| `npx jest --runInBand` (full suite, with the fix) | 918 passed / 920 |
| `npx jest --runInBand` (baseline, changes stashed) | 903 passed / 905 |
| New spec with the fix | 15 / 15 pass |
| New spec with the component changes reverted | 15 / 15 fail |
| `npx eslint .` | 0 errors |
| `npx stylelint "src/**/*.scss"` | clean |
| `npx prettier . --check` | clean |
| `npx htmlhint "src"` | 769 files, 0 errors |
| `node scripts/check-file-headers.js` | all headers valid |
| `npm run build:prod` | success |

The two failures in both runs are the same pre-existing ones in `clients.component.spec.ts` (IME composition, post-destroy debounce). They are present on an unmodified `dev` and are untouched by this change. Exactly the 15 new tests separate the two totals.

**Manual verification against a live Fineract instance was not performed** — no instance was available. The evidence here is the automated suite, the production build, and the Client Notes recording above.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved note creation, editing, and deletion across client, group, center, loan, and savings account views.
  * Note changes now appear immediately and reliably in the interface.

* **Tests**
  * Added regression coverage confirming note updates render correctly after asynchronous operations across all supported note tabs.

### DIFF
--- a/src/app/centers/centers-view/notes-tab/notes-tab.component.ts
+++ b/src/app/centers/centers-view/notes-tab/notes-tab.component.ts
@@ -6,7 +6,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { ChangeDetectionStrategy, Component, OnInit, inject } from '@angular/core';
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, OnInit, inject } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 
 import { AuthenticationService } from '../../../core/authentication/authentication.service';
@@ -28,6 +28,7 @@ export class NotesTabComponent implements OnInit {
   private route = inject(ActivatedRoute);
   private authenticationService = inject(AuthenticationService);
   private centersService = inject(CentersService);
+  private cdr = inject(ChangeDetectorRef);
 
   entityId: string;
   username: string;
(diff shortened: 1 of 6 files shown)


parth-sharma-10:
Superseded by #3858, which carries the Jira key in the branch name and title per `CONTRIBUTING.md` (Steps 3 and 7) and links WEB-1132 / WEB-1133 / WEB-1134. Same commit contents, plus the review feedback from this PR applied. Closing this one so there is a single PR for the change.

PR #3866, opened 2026-08-18
WEB-1149: Implement JLG loan product template

YousufFFFF:
## Description

Adds the JLG (Joint Liability Group) guided product template to the loan product wizard, following the `JLG L` sheet of the loan product re-design workbook.

Unlike every other sheet in the workbook, JLG marks the "Terms vary based on loan cycle" rows applicable, so this adds a new wizard step that lets an operator define per-cycle bands for principal, number of repayments and nominal interest rate. A group member's entitlement is expected to grow with each successfully repaid cycle, and this is how that progression is configured at product level. The step reproduces the equivalent block from the Classic Terms step - same dialog fields, same value-condition options, same emitted object shape - so both flows submit an identical payload; Classic's version is bound directly to its own form group and could not be reused as-is.

The step also validates Fineract's rules for variation lists before submitting, rather than letting the operator discover them through a sequence of 400s: a list must start with an `equals` row, end with a `greater than` row, and carry cycle numbers that advance (`equals` rows run 1, 2, 3…, while a `greater than` row repeats the number above it, since that number is an open-ended threshold). Violations are reported inline against the offending table, reusing Fineract's own already-translated messages. Values are submitted as numbers rather than the strings the shared form dialog produces, because Fineract reads the `locale` for a nested array element from that element itself - which these objects do not carry - so string numerics would otherwise arrive unparsed.

Beyond the new step, the profile seeds both borrower-cycle toggles on, pins the down payment off (the group guarantee is the security), and omits the multi-disbursal tranche family.

Also fixes the product summary view, where the borrower-cycle variation tables rendered with their header row beside the data rather than above it: the tables carried `.flex-fill`, which sets `display: flex` and so discards table layout. Four translation keys in the same view referenced names that do not exist (a casing mismatch and three `(%)` suffix mismatches) and rendered as raw keys; they now point at the keys already present in every locale. Both are pre-existing and were surfaced by JLG being the first guided product to produce loan cycle variations.

Adds the JLG heading, card description and three validation strings across all 13 locales. Every other label the new step needs already existed from the Classic flow.

No dependencies required.

# WEB-1149

## Screenshots, if any
# After Video:

https://github.com/user-attachments/assets/49ac5cc7-5f19-4069-8dc2-f436e211eaae

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Joint Liability Group (JLG) loan product creation.
  * Added borrower-cycle variations for principal, repayment count, and interest rate.
  * Added controls to add, edit, delete, and review cycle variations.
  * Added responsive tables, empty states, and contextual validation messages.
* **Bug Fixes**
  * Prevented submission when borrower-cycle sequences or values are invalid.
  * Ensured disabled borrower-cycle settings submit empty variation data.
* **Localization**
  * Added localized JLG labels, descriptions, and validation messages across supported languages.


YousufFFFF:
@IOhacker This PR is ready for your review!
Thankyou!

IOhacker (approved):

Issue #107, opened 2018-07-06
use fetch client image api in clients layout


anweshknayak:
- clients/view/

- using https://demo.openmf.org/api-docs/apiLive.htm#client_images

Currently the image is harcoded

vatsal3333:
I would like to work on this issue.
@anweshknayak The link which you mentioned above is has an example in it which is showing error when opened. I used the api in the project by adding a getClientImage function in client service file and then used that function in view-client component
![screenshot from 2019-01-05 23-34-38](https://user-images.githubusercontent.com/29050693/50728640-30313600-1153-11e9-93f0-fe234d73b232.png)
![screenshot from 2019-01-05 23-34-28](https://user-images.githubusercontent.com/29050693/50728642-39220780-1153-11e9-9cb2-a06220697032.png)

 and I am getting the same error. Can you please help?

abhaychawla:
Please explore and try to understand the error, the image has not been uploaded so how can it be fetched? You need to handle this case accordingly. Please see the community app for reference to implement the features.

aashish-ak:
@abhaychawla I would like to work on this issue but the clients service, instead of HttpClient uses Http module which is deprecated, so should i use that module?

aashish-ak:
@abhaychawla A little help needed here!

abhaychawla:
@aashish-ak You need to make any relevant change necessary to implement this only using the latest implementation, without making any change in the current code which might be obsolete.

Jov03:
issue already resolved via PR #414

PR #3889, opened 2026-08-22
feat(copilot): name the account on confirmation cards, and settle on one product name

shubhamkumar9199:
## What this changes

Two things a reviewer called out, both fair.

### The confirmation card spoke database

The card was rendering the tool's own arguments, so an officer about to
approve money read this:

```
loanId              12
approvedLoanAmount  28000
```

The gateway now sends labelled, formatted rows alongside the raw arguments,
and the card renders those:

```
Client            Aisha Bello
Loan account      000000000012
Product           Agriculture Term Loan
Applied for       NGN 30,000.00
Approved amount   NGN 28,000.00
Approval date     21 August 2026
```

The raw arguments stay on the wire as the record of what will execute, and
the card still falls back to them, humanised, if it ever meets a gateway too
old to send rows. The idempotency key is shown as a short reference rather
than a full UUID; the gateway keeps the whole thing for the audit trail.

### It had three different names

"Mifos Intelligence AI" in the header, "AI Assistant" on the launcher, "Mifos
AI assistant" on the avatar. It is **Mifos X Copilot** everywhere now, across
all thirteen locales.

Fineract no longer appears in anything an officer reads either. It is what
the platform runs on, not the word a loan officer uses.

## Also in this pass

- Sentence case throughout, rather than a mix of Title Case and sentence
  case in the same menu.
- The starter prompts no longer quote record numbers at people: "What is the
  repayment schedule for this loan?" rather than "for loan #107". Those
  prompts are the first thing anyone sees, and they were teaching the wrong
  vocabulary.
- Copy rewritten to say what it means. "Copilot helps you find answers.
  Anything that moves money still needs your approval." replaces "designed to
  assist, not replace, human decision making."
- Demo mode held to the same bar, since it is what anyone evaluating this
  will actually see.

## Tests

60 to 72. New cases cover row parsing, the gateway's row ordering being
preserved, and blank rows being dropped rather than rendered as empty lines.

## Requires

The gateway side is openMF/mcp-mifosx#434. This is backward compatible: a
gateway that sends no rows still produces a readable card.

## Summary by CodeRabbit

* **New Features**
  * Confirmation cards now show clearer, human-readable action details with formatted values and concise references.
  * Blank or unavailable action details are filtered from displays.
  * Copilot cards, confirmation labels, guidance, and demo content are localized across supported languages.
  * Demo loan and client information uses meaningful labels, titles, and formatted amounts.
  * Recent Chats refresh when opened, with improved account isolation and confirmation-card scrolling.

* **Documentation**
  * Refined Copilot guidance and safety messaging across supported languages.

* **Tests**
  * Added coverage for action-detail ordering, filtering, approval flows, and session isolation.

### DIFF
--- a/src/app/copilot/components/action-card/action-card.component.ts
+++ b/src/app/copilot/components/action-card/action-card.component.ts
@@ -6,10 +6,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { Component, EventEmitter, Input, Output, inject } from '@angular/core';
 import { CommonModule } from '@angular/common';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslateModule, TranslateService } from '@ngx-translate/core';
 import { ActionCard, ActionCardType } from '../../core/models/action-card.model';
+import { translateCardLabel } from '../../core/card-label';
 
 /** Renders a single structured ActionCard (client / loan / savings / insight / confirmation). */
 @Component({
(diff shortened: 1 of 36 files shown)


IOhacker:
@shubhamkumar9199 there are some comments

IOhacker (approved):

PR #3892, opened 2026-08-23
feat(copilot): send the officer's office to the gateway

shubhamkumar9199:
Adds two fields to the context the Copilot panel sends with every turn: `officeId` and `staffId`. Nothing else changes.

The gateway used to create every client with `"officeId": 1` written into its tool manifest as a literal, so an officer in any branch other than head office created clients in the wrong branch, and nothing ever showed it because a test system with one office is right by accident. openMF/mcp-mifosx#438 replaces that literal with the officer's own office, which it reads from the session context this panel sends, and this panel was not sending it. `Credentials` already carries both fields from the authentication response, so nothing new is fetched.

Sending it beside the conversation rather than through it is the point. An office is a fact about who is signed in, not a decision. If it went through the language model then a branch could be chosen by a sentence, and a client could be created in someone else's branch by asking for it nicely. The model never sees the value.

Safe to merge in either order. Without the gateway change these fields are ignored; with the gateway change and without this one, client creation refuses with a message naming the missing detail rather than quietly posting a client with no office.

89 tests, unchanged and passing.

## Summary by CodeRabbit

* **New Features**
  * Copilot context now includes the signed-in officer’s office and optional staff identifiers.
  * Context values default safely when unavailable, improving support for personalized Copilot interactions.

* **Tests**
  * Updated Copilot-related test scenarios to validate office and staff context information.

### DIFF
--- a/src/app/copilot/core/mcp-client.spec.ts
+++ b/src/app/copilot/core/mcp-client.spec.ts
@@ -25,6 +25,8 @@ const REQUEST: McpChatRequest = {
     loanId: null,
     screen: 'client-detail',
     loggedInUser: 'priya',
+    officeId: 4,
+    staffId: null,
     role: 'Loan Officer',
     language: 'en'
   }
--- a/src/app/copilot/core/models/copilot-context.model.ts
+++ b/src/app/copilot/core/models/copilot-context.model.ts
@@ -22,6 +22,14 @@ export interface CopilotContext {
   screen: string;
   /** Username decoded from the JWT. */
   loggedInUser: string;
+  /**
+   * The office the signed-in officer belongs to. Sent so the gateway can put a new client in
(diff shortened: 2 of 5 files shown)


shubhamkumar9199:
Closing this: the approach changed under review.

This added officeId and staffId to the context the panel sends, so the gateway could stop hardcoding `"officeId": 1`. Review on openMF/mcp-mifosx#438 pointed out that a value the server can establish for itself should not arrive on trust from a client, and that was the right call.

The gateway now asks Fineract which offices the officer's own forwarded credential can reach, and uses that. Nothing about their identity comes from the request body, so these fields have no reader and would be dead weight in the context payload.

Verified against the sandbox with a real model: the panel sends no office, the gateway works it out, and the client is recorded in the officer's own office.

PR #3917, opened 2026-08-26
WEB-1184: [CI/CD] Add a nightly Fineract drift canary

devvaansh:
## Description

WEB-1180 (#3914) pins Fineract to an immutable digest, which fixes reproducibility — but a pin defers upgrade pain rather than removing it. Nothing notices when the pinned build falls behind upstream, so the cost surfaces all at once at the next manual bump, including the exact defect class that motivated the pin (a stretchy report shipped without its `stretchy_report_parameter` row, which surfaces as a 403 → failed route resolver → redirect to `#/`, nowhere near the spec that goes red).

This adds `.github/workflows/e2e-nightly.yml` — a scheduled job that deliberately runs the E2E suite against a **moving** tag, in a job that is allowed to fail, so drift is caught continuously instead of by accident inside someone's PR.

**Two signals from one run:**

1. **Upstream drift** — fails here but passes on the pinned digest ⇒ Fineract changed under us and the next bump will break.
2. **A real flake** — fails on both ⇒ `Playwright E2E`'s `--retries=2` has been masking it.

To make signal 2 meaningful the canary runs with `--retries=0`. Retries exist to keep PRs unblocked; here they'd hide the thing we're looking for.

**Design notes worth reviewing:**

- **Moving tag via `FINERACT_IMAGE`.** `docker-compose.e2e.yml` reads `${FINERACT_IMAGE:-<pinned digest>}` since WEB-1180, so the canary only sets that variable (default `apache/fineract:latest`); the pin used by `Playwright E2E` is untouched.
- **Separate concurrency group.** `playwright.yml` uses `e2e-${{ github.ref }}` with `cancel-in-progress: true`. A scheduled run and a push to `dev` both resolve to `refs/heads/dev`, so sharing that group would let the canary cancel the required `Playwright E2E` check — or be cancelled by it. This uses `group: e2e-nightly`, `cancel-in-progress: false`.
- **Records the resolved digest before running.** The point of a canary is knowing *which* upstream build broke us, so the digest goes into the job summary up front.
- **Same boot path as `Playwright E2E`.** A canary that boots differently from the real job produces readiness failures that look like drift. Following WEB-1218 (#3968), it relies on the same readiness contract — `docker compose up -d` waits on the compose healthchecks and `playwright/global-setup.ts` probes seed data and the web-app — with no workflow-level wait steps of its own. It runs the same `npm run playwright:ci` command (plus `--retries=0`), so reporters stay in sync.
- **No flaky-test summary step.** The WEB-1238 (#3991) summary reports tests that passed on retry; with retries off there is nothing for it to report, so it is left out.
- `workflow_dispatch` takes a `fineract_image` input so the override can be exercised manually rather than waiting for 02:00 UTC.

WEB-1184

Builds on WEB-1180 (#3914), WEB-1218 (#3968) and WEB-1238 (#3991), all merged.

## Screenshots, if any

N/A — CI workflow addition, no UI surface.

## Checklist

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -1,0 +1,180 @@
+name: Cron – Nightly Fineract Drift Canary
+
+on:
+  schedule:
+    # 02:00 UTC daily. GitHub delays scheduled runs under load and skips
+    # them on inactive repos; this canary has no deadline, so a late or
+    # missed run costs nothing.
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+    inputs:
+      fineract_image:
+        description: 'Fineract image to run the suite against'
+        required: true
+        default: 'apache/fineract:latest'
+        type: string
+
+permissions:
+  contents: read
+
+# Deliberately NOT the `e2e-${{ github.ref }}` group used by playwright.yml.
+# A scheduled run and a push to dev both resolve to refs/heads/dev, so a
+# shared group with cancel-in-progress would let this canary cancel the
+# required `Playwright E2E` check — or be cancelled by it.
+concurrency:
+  group: e2e-nightly
+  cancel-in-progress: false
+
+jobs:
+  canary:
+    name: E2E Drift Canary
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    env:
+      CI: true
+      # Overrides the digest pinned in docker-compose.e2e.yml (WEB-1180) so
+      # this job deliberately tracks a moving tag. Until WEB-1180 lands the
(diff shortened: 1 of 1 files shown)


IOhacker:
@devvaansh tell us if there is any dependency or blocker in order to be solved for moving the PR to ready for review

devvaansh:
>@devvaansh tell us if there is any dependency or blocker in order to be solved for moving the PR to ready for review.

@IOhacker  still blocked on the dependency PR , it isn't open yet, I'll put it up this weekend and flip this to ready once it's in.

devvaansh:
@IOhacker now ready for review

IOhacker (approved):

PR #3925, opened 2026-08-28
WEB-1189: Repair search spec selectors for IME regression test

YousufFFFF:
## Description

Two tests in `clients.component.spec.ts` fail on committed `dev`, leaving the suite red for everyone.

The spec queries `input[matInput]`, but the client list redesign replaced the `mat-form-field` search box with a plain input. The selector matches nothing, so the test throws on a null element before reaching any assertion — meaning the IME composition behaviour it was written to protect has been uncovered ever since. It now selects the input by its container.

The destroy test called `component.ngOnDestroy()` and expected the pending debounce to be cancelled. Teardown runs through `takeUntilDestroyed(destroyRef)`, which fires on real component destruction, not on a manual lifecycle call, so the test asserted a contract the component never promised. It now uses `fixture.destroy()` to exercise the real path.

Both changes are test-side. No production behaviour is affected.

# WEB-1189
## Screenshots, if any
NA
## Checklist

## Summary by CodeRabbit

* **Tests**
  * Updated client search test coverage to reflect the current search input behavior.
  * Added fixture cleanup when testing cancellation of pending searches.

### DIFF
--- a/src/app/clients/clients.component.spec.ts
+++ b/src/app/clients/clients.component.spec.ts
@@ -116,7 +116,7 @@ describe('ClientsComponent — debounce search', () => {
   });
 
   it('should not search during IME composition but should search on compositionend', () => {
-    const inputEl: HTMLInputElement = fixture.nativeElement.querySelector('input[matInput]');
+    const inputEl: HTMLInputElement = fixture.nativeElement.querySelector('.search input');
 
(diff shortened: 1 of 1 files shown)


YousufFFFF:
Hello @IOhacker!
Can you review this PR?

IOhacker (dismissed):

IOhacker (approved):

PR #3926, opened 2026-08-28
WEB-1191: Show full client name in the clients list

AnshikaSingh0011:
## Description
On the Clients list page, long client names were being cut off with an ellipsis and there was no way to read the full name without opening the client. For example a name like "JHOAN SEBASTIAN BECERRA RODRIGUEZ" showed up as "JHOAN SEBASTIAN BECERRA RODRIG...", so it is hard to tell apart two clients whose names start the same way.

There were two reasons for this:

- The NAME column had a maximum width of 32%, and the leftover space in the row was shared equally between all the columns. So the account number / external id / office columns were taking more space than they need while the name column stayed small even on a big screen.
- .name used white-space: nowrap with text-overflow: ellipsis, so anything that did not fit was simply hidden.

## Changes Made
- The name column now uses minmax(220px, 1fr) so it takes whatever space the other columns don't need. I also lowered the other columns from 16%/16%/18% to 14%/14%/16%, they still have enough room for their content.
- .name now wraps to a second line instead of hiding the text behind an ellipsis (-webkit-line-clamp: 2), so on smaller screens the full name is still readable.
- Added a title tooltip on the name so hovering shows the complete name. For the clients list the tooltip uses the same masked value as the row when complianceHideClientData is on, so nothing is leaked.

The Groups and Centers lists use the same _entity-list.scss styles, so they get the same fix.

https://mifosforge.jira.com/browse/WEB-1191

## Screenshots, if any

before:
client name issue

after:
{D0E82B33-2747-4E13-BA38-DA5B41B07176}

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved display of long center and group names with hover tooltips.
  - Long names now wrap across up to two lines instead of being cut off.
  - Adjusted list column spacing to provide more room for names.
  - Improved client name display while preserving configured privacy masking.


src/app/clients/clients.component.html
```diff
@@ -103,7 +103,7 @@
           {{ initials(row.displayName) }}
         </div>
         <div class="name-block">
-          <span class="name">{{ hideClientData ? maskName(row.displayName) : row.displayName }}</span>
```
IOhacker, line 106: the mask name is required for modes where the data must not be displayed. This feature is available?

src/app/clients/clients.component.ts
```diff
@@ -74,6 +74,12 @@ export class ClientsComponent implements OnInit, OnDestroy {
       .join(' ');
   }
 
+  /** Name as shown in the list, masked when client data must be hidden. */
+  displayedName(row: any): string {
+    const name: string = row?.displayName ?? '';
+    return this.hideClientData ? this.maskName(name) : name;
```
IOhacker, line 80: I think it solves my previous comment

IOhacker (approved):

src/app/clients/clients.component.ts
IOhacker: solved

PR #3931, opened 2026-08-30
WEB-1195: Add tooltips and accessible names to two toolbar buttons

AnshikaSingh0011:
## Description

The theme toggle and the user profile button in the main toolbar don't show a
tooltip on hover, and they have no accessible name either, so a screen reader
just reads out "button" for both. They seem to be the only two controls in that
toolbar missing this. I've added a tooltip and an aria-label to each.

For the theme toggle the label changes with the theme, since the icon shows the
theme you'll get after clicking and not the one you're currently in. I put the
ternary in a getter instead of the template so the tooltip and the aria-label
stay in sync.

While fixing the profile button I noticed its alt text was bound to
`labels.headings.User Profile`, but `en-US.json` only has `labels.heading`
(singular), so the key was never resolving and the raw key was being read out.
I've corrected it. It was the only place in `src` using the plural spelling.

The two new tooltip strings are added to `en-US.json` for translation.

https://mifosforge.jira.com/browse/WEB-1195

## Checklist

## Summary by CodeRabbit

* **Accessibility**
  * Added accessible labels and helpful tooltips to the user profile and theme toggle controls.
  * Corrected the user profile image’s alternative text translation for improved screen reader support.
* **Localization**
  * Added theme-switching tooltip translations across supported languages.
  * Added localized text for report uploads, onboarding, tasks, validation messages, and related workflows in several languages.
* **User Experience**
  * Theme toggle tooltips now clearly indicate whether selecting the control will switch to light or dark mode.

### DIFF
--- a/src/app/core/shell/toolbar/toolbar.component.html
+++ b/src/app/core/shell/toolbar/toolbar.component.html
@@ -124,8 +124,15 @@
   </div>
 
   <div #appMenu>
-    <button mat-icon-button class="ml-1 img-button" [matMenuTriggerFor]="applicationMenu" tabindex="0">
-      <img src="assets/images/user_placeholder.png" [alt]="'labels.headings.User Profile' | translate" />
+    <button
+      mat-icon-button
+      class="ml-1 img-button"
+      [matMenuTriggerFor]="applicationMenu"
+      tabindex="0"
+      [attr.aria-label]="'labels.heading.User Profile' | translate"
+      matTooltip="{{ 'labels.heading.User Profile' | translate }}"
+    >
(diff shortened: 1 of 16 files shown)


IOhacker (changes requested):
@AnshikaSingh0011 could you please add the translation for the other i18n languages ?

AnshikaSingh0011:
> @AnshikaSingh0011 could you please add the translation for the other i18n languages ?

thanks for pointing that out . have made the changes also I'm not too sure about the Nepali and Swahili translation . Could
someone who speaks either have a look before this gets merged?

IOhacker (approved):

PR #3946, opened 2026-09-02
WEB-1209: Explain why the guided loan product wizard cannot submit

YousufFFFF:
## Description

In the guided (template) loan product wizard, "Create Loan Product" was a silent dead button. On an invalid form `submit()` called `markAllAsTouched()` and returned — no message, no stepper move, no disabled state. Because the stepper is `[linear]="false"`, the operator reaches Review with an earlier step still incomplete, so the control that blocked the create is usually on a step they are not looking at and the revealed field messages are off screen. The user clicks Create, nothing happens, repeatedly, with no way to discover why.

Classic mode in the same component already solves this with `incompleteClassicSteps` and its `review-pending` block. This ports that affordance to the guided path — the one aimed at less expert operators, and the one that did not have it.

A failed create now renders a `role="alert"` summary on the Review step naming every step that still blocks it, each a button that jumps straight to that step. The summary appears on the click rather than at page load, and clears itself as the last offending control is fixed.

The live region also closes the accessibility side of the same defect: `markAllAsTouched()` reveals messages silently, so a screen-reader user previously got no signal at all that the click had been rejected.

Notable decisions:

- `guidedSubmitBlocked` is one expression read by both `submit()` and the template, so the gate and the explanation cannot drift apart.
- `incompleteGuidedSteps` maps the flat guided FormGroup back to steps through `visibleFields`, the same visibility source of truth the Review summary uses, so a field the operator was never shown can never be named.
- The stepper is deliberately not auto-moved on a failed submit — that would pull the operator off the summary they just asked for. The per-step buttons give them the jump instead.
- `guidedSubmitBlockedWithoutCause` renders a generic line if the walk ever finds no owning step, so a blocked submit is always explained.

Two new translation keys, added to all 13 locales.

# WEB-1209

## Screenshots, if any

WhatsApp Image 2026-09-02 at 10 43 18 PM

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added clearer guidance when creating a loan product is blocked by incomplete or invalid steps.
  - Review now lists incomplete steps as clickable links to the relevant step.
  - Added a general prompt to check highlighted fields when no specific incomplete step is identified.
  - Added accessible alert messaging for blocked submissions.

- **Bug Fixes**
  - Prevented overlapping warnings when the incomplete-step summary is displayed.

- **Localization**
  - Added product-creation guidance translations across supported languages.

- **Tests**
  - Added coverage for blocked submissions, navigation, accessibility announcements, and completion updates.

### DIFF
--- a/src/app/products/loan-products/wizard/loan-product-wizard.component.html
+++ b/src/app/products/loan-products/wizard/loan-product-wizard.component.html
@@ -326,9 +326,52 @@
           </div>
         </ng-container>
 
-        <p *ngIf="i === visibleSteps.length - 1 && borrowerCycleStepInvalid" class="step-actions__blocked">
+        <!--
+          Suppressed once the summary below is up: the loan-cycle step is one of the entries it lists,
+          and two messages stacked on the same problem read as two problems. This one stays for the
+          pre-attempt case, where it is the only warning the operator gets.
+        -->
+        <p
+          *ngIf="i === visibleSteps.length - 1 && borrowerCycleStepInvalid && !showGuidedIncompleteSummary"
+          class="step-actions__blocked"
+        >
           {{ 'labels.text.Fix the loan cycle variations before creating the product' | translate }}
         </p>
+
+        <!--
+          C4: the answer to a "Create Loan Product" click that cannot go through. `role="alert"` is not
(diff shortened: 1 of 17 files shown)


YousufFFFF:
@IOhacker Please review this PR!

IOhacker (approved):

PR #3793, opened 2026-08-04
WEB-1101: Add missing translation keys for Disable/Enable Withhold Ta…

Farah-Nahle-FOO:
## Description

The "More" submenu on the savings account hamburger menu displayed the raw i18n key `labels.menus.Disable Withhold Tax` (and `Enable Withhold Tax`) instead of a readable label, because the `menus` translation namespace never had entries for these two menu actions. Added the missing `Disable Withhold Tax` and `Enable Withhold Tax` keys to `src/assets/translations/en-US.json` so `ngx-translate` resolves a proper label instead of falling back to the raw key. This also fixes the same menu action on the recurring deposit and fixed deposit account views, since they share the same translation keys. No new dependencies are required.

## Related issues and discussion

[WEB-1101](https://mifosforge.jira.com/browse/WEB-1101)

image

_Before:_ menu item shows `labels.menus.Disable Withhold Tax`
_After:_ menu item shows `Disable Withhold Tax`

## Checklist

Please make sure these boxes are checked before submitting your pull request - thanks!

- [x] If you have multiple commits please combine them into one commit by squashing them.

- [x] Read and understood the contribution guidelines at `web-app/.github/CONTRIBUTING.md`.

[WEB-1101]: https://mifosforge.jira.com/browse/WEB-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added English menu labels for enabling and disabling withholding tax.

PR #3996, opened 2026-09-12
WEB-1101: Translate the withhold tax and annual fee account menu items

parth-sharma-10:
## Description

The savings account **More** submenu currently shows `labels.menus.Disable Withhold Tax` instead of a translated label.

The savings, fixed deposit, recurring deposit, and shares account menus all use the option's `name` as the translation key. The savings and recurring deposit templates build the key with `labels.menus.` + `option.name`, while the fixed deposit and shares templates use `translateKey: 'menus'`, which resolves to the same namespace.

The problem is that `Disable Withhold Tax` and `Enable Withhold Tax` only exist under `labels.heading`, which is what the confirmation dialogs use. They are missing from `labels.menus`, so the menu displays the raw translation key.

While checking all 36 option names across these four menus, I found one more missing key: **Apply Annual Fees**, used by the savings and recurring deposit menus. I’ve included that as well rather than fixing only the two labels from the ticket.

### Locales

The fix covers all 13 supported locales, not just en-US.

The app uses `TranslateService.use()` but never calls `setDefaultLang()`, so a missing translation does not fall back to English. `CustomMissingTranslationHandler` only unwraps `labels.catalogs.`; other missing namespaces are displayed as-is.

For the withhold-tax labels, I reused the wording already present under each locale's `labels.heading`, so the menu and confirmation dialog use the same text. For **Apply Annual Fees**, the wording is composed from terms that already exist in each locale.

### Shares typo

The shares account view had the option written as `Apply Anuual Fees`. I corrected the typo so the same translation key can be shared across all three account types.

The shares `doAction` still has no case for this option, so the menu entry remains inert. That is a separate issue and isn't addressed here.

### Test

Added `web-1101-account-menu-translations.spec.ts`.

The test checks the three affected keys in every locale and also verifies that every option declared by the four account menus has an en-US translation.

The test reads the account view sources directly because the two tax options aren't in a buttons configuration. They are added by the account view components when the relevant `taxGroup` / `charges` conditions are met.

Without the fix, 14 of the 15 test cases fail.

### Other missing translations

The same sweep also found some pre-existing missing translations that are outside the scope of this change:

* `Interbank Transfer` and `Reporting Dashboard` are missing from `labels.menus` in all 12 non-English locales.
* `Assign Staff` is missing in it-IT, lv-LV, and sw-SW.
* `Recovery Payment` is missing in ko-KO.

These need actual locale-specific translations rather than reusing an existing string, so they should be handled separately as part of a locale translation pass.

## Related issues

[[WEB-1101](https://mifosforge.jira.com/browse/WEB-1101)](https://mifosforge.jira.com/browse/WEB-1101)

This continues #3793, which has been inactive for some time. The underlying issue is the same, but this change also covers the other locales, the additional missing menu entry, and regression coverage.

## Checklist

* [x] Squashed all changes into one commit.
* [x] Read and understood `web-app/.github/CONTRIBUTING.md`.

[WEB-1101]: https://mifosforge.jira.com/browse/WEB-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the “Apply Annual Fees” button label.
  * Improved consistency and accuracy of withholding-tax terminology in several languages.

* **Localization**
  * Added translated menu labels for annual fees and withholding-tax actions across supported locales.
  * Updated related headings to match established terminology.

* **Tests**
  * Added coverage verifying account-menu translations across all supported locales and identifying missing labels.

### DIFF
--- a/src/app/core/translation/web-1101-account-menu-translations.spec.ts
+++ b/src/app/core/translation/web-1101-account-menu-translations.spec.ts
@@ -1,0 +1,129 @@
+/**
+ * Copyright since 2025 Mifos Initiative
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { CustomMissingTranslationHandler } from './missing-translation.handler';
+
+/**
+ * WEB-1101: the savings, fixed deposit, recurring deposit and shares account menus render every
+ * entry as `labels.menus.<option name>` — the savings and recurring deposit templates concatenate
+ * the prefix, the fixed deposit and shares ones go through `translateKey: 'menus'`, which builds
+ * the same key. The option name is therefore an implicit translation key, and a name with no entry
+ * under `labels.menus` reaches the operator as the raw key: `labels.menus.Disable Withhold Tax`.
+ *
+ * Nothing catches that at build time, and no locale falls back to English — the app only ever calls
+ * `TranslateService.use()`, never `setDefaultLang` — so a missing key is visible in all 13 locales.
+ */
+
+const locales = [
+  'cs-CS',
+  'de-DE',
+  'en-US',
+  'es-CL',
+  'es-MX',
+  'fr-FR',
+  'it-IT',
+  'ko-KO',
+  'lt-LT',
+  'lv-LV',
+  'ne-NE',
+  'pt-PT',
+  'sw-SW'
+];
+
+/** The menu names WEB-1101 added. Kept explicit so a locale dropping one is a named failure. */
+const web1101MenuNames = [
+  'Apply Annual Fees',
+  'Disable Withhold Tax',
+  'Enable Withhold Tax'
+];
+
+/** Every file that names an entry of those four account menus. */
+const menuSources = [
+  'src/app/savings/savings-account-view/savings-buttons.config.ts',
+  'src/app/savings/savings-account-view/savings-account-view.component.ts',
+  'src/app/deposits/recurring-deposits/recurring-deposits-account-view/recurring-deposits-buttons.config.ts',
+  'src/app/deposits/recurring-deposits/recurring-deposits-account-view/recurring-deposits-account-view.component.ts',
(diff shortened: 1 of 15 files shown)


parth-sharma-10:
@coderabbitai thanks — holding all nine of the wording suggestions, with a reason and a proposal.

Every string flagged is copied verbatim from that locale's existing `labels.heading.Disable Withhold Tax` / `Enable Withhold Tax`, which is what the confirmation dialog the menu item opens already renders (`recurring-deposits-account-view.component.ts:315,339`). That was deliberate: taking the suggestions here would make the menu and the dialog it opens say different things in nine locales, which is a worse defect than the wording.

The suggestions are also not converging on one term. Checked each against the noun the same file already uses for `Withhold Tax` in `labels.catalogs`:

| locale | file's own term | suggested |
|---|---|---|
| cs-CS | Srážková daň | srážkovou daň (matches) |
| lv-LV | Ieturējuma nodoklis | ieturējuma nodokli (matches) |
| sw-SW | Kodi ya Kuzuia | Kodi ya Kuzuia (matches) |
| fr-FR | Retenir l'impôt | retenue à la source (third term) |
| it-IT | Trattenuta fiscale | ritenuta d'acconto (third term) |
| lt-LT | Išskaičiuoti mokestį | išskaičiuojamąjį mokestį (third term) |

So three of them would align the menu with the catalogs noun while breaking it against the heading, and three would introduce a term the file does not use anywhere. Same for it-IT `Apply Annual Fees`: `tariffe annuali` is the file's own phrasing (`labels.inputs.Annual Fees Transaction Date`), `commissioni annuali` is new.

The wording is worth fixing — the pt-PT heading (`Ativar imposto reter`) is not idiomatic and neither is the ne-NE one — but it is a change to strings that predate this PR, and to be coherent it has to move `labels.heading`, `labels.menus` and `labels.catalogs` together, per locale, with native-speaker review. This PR is a missing-key fix; I would rather not smuggle a partial retranslation into it. I will raise it as a locale pass alongside the gaps the same sweep found (`Interbank Transfer` and `Reporting Dashboard` missing from `labels.menus` in all 12 non-English locales, `Assign Staff` in it-IT/lv-LV/sw-SW, `Recovery Payment` in ko-KO), and tag native speakers there.

Happy to be overruled by a maintainer or by anyone who actually speaks the language in question.

IOhacker (approved):

PR #4004, opened 2026-09-14
WEB-1244: derive the report engine chips from the loaded reports

AnshikaSingh0011:
## Problem
On the Reports Catalog the engine filter counts do not add up to the total shown in the heading. On a standard Fineract instance the heading reports 84 available reports, while the chips show Table 59, Pentaho 0, BIRT 0, Chart 0 and SMS 22, which total 81.

The three missing reports are of type Email (Loan Approved - Email, Loan Rejected - Email, Loan Repayment - Email). Since no chip corresponds to that type they appear only under All, and selecting any engine hides them.

The same fixed list also produced chips for engines that no report uses. Pentaho, BIRT and Chart were each shown with a count of zero, and selecting one produced an empty catalogue.

### Changes
- Build the engine chips from the report types actually present in the loaded
  data, so every report belongs to a chip and the counts add up to the total.
- Keep the familiar engines first by ordering against a preferred list, and
  append any remaining type alphabetically.
- Add reports.component.spec.ts covering the total, an engine outside the
  preferred list, the exclusion of unused engines, the ordering, and filtering
  by a derived engine.

https://mifosforge.jira.com/browse/WEB-1244

## Screenshots, if any:
before:
{ADB1205B-22DA-4565-BD34-C2E6D92A6D02}
email before the changes shown by filter which isn't shown in table:
{4577D1E9-37BA-44CE-A897-4FB46B502BB9}

after:
{4CF53871-D6D2-430E-AD8B-431FB1C70951}

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Report engine filters now include every engine represented in the loaded reports.
  - Engine filters are displayed in a preferred order, with additional engines listed alphabetically.
  - Each engine filter shows its corresponding report count.
  - Engines with no reports are omitted from the filter list.

- **Tests**
  - Added coverage for engine counts, ordering, filtering, and handling of previously unlisted engines.


IOhacker (changes requested):
@AnshikaSingh0011 have you tested with the Pentaho and the Eclipse Birt plugins deployed?

AnshikaSingh0011:
> @AnshikaSingh0011 have you tested with the Pentaho and the Eclipse Birt plugins deployed

@IOhacker No , I haven't tested it with pentaho and eclipse brt plugins deployed , my setup is the stock Fineract image with only Table, SMS and Email reports. The chips are now built from the report types the API returns instead of a fixed list, so a
deployment with Pentaho or BIRT reports will still show those chips, and I have added a spec covering that. What changes is that an engine with no reports no longer gets a chip, which is why Pentaho, BIRT and Chart disappear on a stock instance where each showed 0. Should a zero count chip still be shown to indicate the engine is supported?

PR #4021, opened 2026-09-17
fix(documents): render PDF previews for client, loan and savings documents

Shauryan0207:
## Description

Previewing a PDF from **Client Documents**, **Client Identifiers**, **Loan Account Documents** or **Savings Account Documents** did nothing — the file downloaded but never rendered. Image (JPG/PNG) previews worked and are unchanged by this PR.

**Root cause.** `DocumentPreviewService` created its object URL directly from the blob Fineract returned, and Fineract serves document attachments as `application/octet-stream`. An `` ignores the declared type and sniffs the bytes, which is exactly why JPG/PNG previews worked — but `<embed>`/`<iframe>` honour it, so the browser offered the PDF as a download instead of displaying it. Type detection also read `blob.type` first, so that generic `octet-stream` value shadowed the reliable `fileName` extension fallback.

PDFs were additionally rendered as lightGallery `iframe` slides with `zoomFromOrigin: true`. lightGallery only supports `zoomFromOrigin` for image slides, and an iframe slide has no fallback on Safari/iOS, which refuse to render `blob:` PDFs inline.

**Changes**

- `document-preview.service.ts` — detect the document type *before* creating the object URL, prefer document metadata over the response `Content-Type`, and re-wrap the blob with a concrete MIME type so `<embed>` renders it. Generic MIME types (`application/octet-stream` and friends) no longer short-circuit detection.
- New shared `PdfPreviewDialogComponent` — renders the PDF with the `DomSanitizer.bypassSecurityTrustResourceUrl` + `<embed type="application/pdf">` pattern already used elsewhere in the app, plus an **Open in new tab** fallback for Safari/iOS. No new dependency.
- `entity-documents-tab.component.ts` (clients, loans and savings documents) and `identities-tab.component.ts` (client identifiers) — route PDFs to that dialog; lightGallery now receives images only and keeps working exactly as before. `zoomFromOrigin` disabled since it is unsupported in dynamic mode.
- Thumbnail prefetching is limited to images, so opening a documents tab no longer downloads every PDF in it just to discard the blob.
- Added `labels.buttons.Open in new tab` and `labels.heading.Document Preview` to `en-US.json`.

**Tests**

- New `document-preview.service.spec.ts` covering the octet-stream PDF and image cases, an already-correct `Content-Type`, preview caching, and classification.
- `entity-documents-tab.component.spec.ts` gains coverage that a PDF opens the dialog rather than the lightbox, and that non-image documents are not prefetched.
- Full suite green: 103 suites / 1276 tests. `npm run build` and lint (eslint, stylelint, prettier, htmlhint) pass.

No dependencies added.

MXWAR-148

## Screenshots, if any

_Not included — this needs a running Fineract instance with uploaded PDF documents to capture. Manual verification steps are in the description above._

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated PDF preview dialog with options to open the document in a new tab or close the preview.
  - Image documents continue to open in the image gallery, while PDFs and other non-image documents are handled separately.
  - Improved document type detection so PDFs and images preview correctly even when server metadata is generic.
  - Added clearer preview titles, labels, and controls.


IOhacker:
@Shauryan0207 there are some conflicts but I think this ticket should target the web-app-react repo https://github.com/openMF/mifos-x-web-app-react. This one already has this feature.

IOhacker (changes requested):
This is the repo https://github.com/openMF/mifos-x-web-app-react

Shauryan0207:
>This is the repo https://github.com/openMF/mifos-x-web-app-react

 closing here and re-raising it against that repo.

